Issue #6, opened 2026-02-19
Add optional build of IAL


uandrae:
Introduce build of https://github.com/ACCORD-NWP/IAL starting with CY50T2 ( or develop ). Currently Cmake compilation fails on atos_bologna, https://github.com/ACCORD-NWP/IAL/issues/561.

trygveasp:
I'll await the hunt for success in https://github.com/ACCORD-NWP/IAL/issues/561 
It's nice to have an example which is supposed to work when implementing. Please ping here if you have success compiling:-)

trygveasp:
Started on a branch https://github.com/trygveasp/tactus/tree/feature/ial_bundle_compilation

Can be started with:
```
# Just use any config (Suite name will be generated based on this)
tactus case ?tactus/data/config_files/configurations/cy49t2_arome tactus/data/config_files/modifications/compile.toml -o config.toml
tactus start suite --config-file config.toml 
```

Tried the commands in BatchJob's, but there are some issues with git command from the tasks. I cloned the CY50T2 repo first (`/perm/sbu/github/ACCORD-NWP/IAL_tags/CY50T2`), but the bundle tool also fails because of assumed git ssh access?

```
+ git clone -o origin https://github.com/ACCORD-NWP/oops /etc/ecmwf/nfs/dh2_perm_b/sbu/github/ACCORD-NWP/IAL_tags/CY50T2/bundle/source/oops
Cloning into '/etc/ecmwf/nfs/dh2_perm_b/sbu/github/ACCORD-NWP/IAL_tags/CY50T2/bundle/source/oops'...
fatal: could not read Username for 'https://github.com': No such device or address
ERROR: Command git clone -o origin https://github.com/ACCORD-NWP/oops /etc/ecmwf/nfs/dh2_perm_b/sbu/github/ACCORD-NWP/IAL_tags/CY50T2/bundle/source/oops failed 
ERROR: Could not download or update oops ...
```

But... it's a start...

uandrae:
For the bundle you export GITHUB as `https://github.com` or `git@github.com:` and I guess you have to be able to authorize yourself in some way?

uandrae:
The develop branch buiilds with changes suggested in https://github.com/ACCORD-NWP/IAL/issues/561.

trygveasp:
The problem is how to reach github from the batch jobs:

```
[...]

PR #11, opened 2026-03-02
Feature/ial bundle compilation

trygveasp:
## Describe your changes

A first version to solve https://github.com/ACCORD-NWP/tactus/issues/6

New settings (use a branch with hacks to compile CY50T2):
```
[compile]
  arch = "ecmwf/hpc2020"
  git_branch = "bugfix/compile_hacks"
  git_repo = "git@github.com:trygveasp/ACCORD-IAL"
```

Activate by (in normal experiments):
```
[suite_control]
compile = true
```

or only a compilation suite:
```
[suite_control]
compile = true
suite_definition = "CompilationSuiteDefinition"
```

image

remaining problem is to actually use the compiled binaries

## Type of change

### Testing

### Code quality

## Checklist for reviewers
- [ ] the code readable
- [ ] the code documented

  missing)

## Checklist for assignees
- [ ] the tests passing

## Tag possible reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/destination-earth-digital-twins/Deode-Prototype/tree/HEAD)
 
 ### Added
+- Add compilation and use of binaries from IAL code built with IAL-bundle [\#11](https://github.com/ACCORD-NWP/tactus/pull/11) (@trygveasp)
 - Support multiple search paths for json schemas. [\#5](https://github.com/ACCORD-NWP/tactus/pull/5) (@mfroelund)
 - Reintroduce remove functionality. [\#7](https://github.com/ACCORD-NWP/tactus/pull/7) (@mfroelund)
 - Support for specifying modification files with subpaths. [\#10](https://github.com/ACCORD-NWP/tactus/pull/10) (@mfroelund)
--- a/tactus/data/config_files/config_file_schemas/main_config_schema.json
+++ b/tactus/data/config_files/config_file_schemas/main_config_schema.json
@@ -632,7 +632,7 @@
         "git_repo": {
             "type": "string",
             "description": "Git repo for IAL source",
-            "default": "https://github.com/ACCORD-NWP/IAL"
+            "default": "git@github.com:ACCORD-NWP/IAL"
         },
         "git_branch": {
             "type": "string",
@@ -644,7 +644,7 @@
           "description": "Compiler configuration",
           "default": "ecmwf/hpc2020"
         }
-      }      
+      }
     },
     "gribmodify": {
       "title": "Section for modify grib",
--- a/tactus/data/config_files/config_file_schemas/main_config_schema.json
+++ b/tactus/data/config_files/config_file_schemas/main_config_schema.json
@@ -634,6 +634,11 @@
             "description": "Git repo for IAL source",
             "default": "git@github.com:ACCORD-NWP/IAL"
         },
+        "git_token": {
+            "type": "string",
+            "description": "Git token for IAL source access",
+            "default": ""
+        },
         "git_branch": {
             "type": "string",
             "description": "Branch/tag",
--- a/tactus/data/config_files/config_file_schemas/main_config_schema.json
+++ b/tactus/data/config_files/config_file_schemas/main_config_schema.json
@@ -637,7 +637,7 @@
         "git_branch": {
             "type": "string",
             "description": "Branch/tag",
-            "default": "CY50T2"
+            "default": "develop"
         },
         "arch": {
           "type": "string",
--- a/tactus/data/config_files/config_file_schemas/main_config_schema.json
+++ b/tactus/data/config_files/config_file_schemas/main_config_schema.json
@@ -17,6 +17,9 @@
     "system": {
       "$ref": "#/definitions/system"
     },
+    "compile": {
+      "$ref": "#/definitions/compile"
+    },
     "boundaries": {
       "$ref": "#/definitions/boundaries"
     },
@@ -367,6 +370,11 @@
           "type": "string",
           "default": "TactusSuiteDefinition"
         },
+        "compile": {
+          "description": "Compile IAL source code",
+          "type": "boolean",
+          "default": false
+        },
         "do_creategrib_static": {
           "description": "Activate grib conversion for static files",
           "type": "boolean",
@@ -611,6 +619,32 @@
           "default": "@CYCLE@/e927.json"
         }
       }
+    },
+    "compile": {
+      "title": "Compilation configuration",
+      "type": "object",
+      "properties": {
+        "ial_dir": {
+            "type": "string",
+            "description": "Local path to IAL source dir",
+            "default": "@CASEDIR@/IAL"
+        },
+        "git_repo": {
+            "type": "string",
+            "description": "Git repo for IAL source",
+            "default": "https://github.com/ACCORD-NWP/IAL"
+        },
+        "git_branch": {
+            "type": "string",
+            "description": "Branch/tag",
+            "default": "CY50T2"
+        },
+        "arch": {
+          "type": "string",
+          "description": "Compiler configuration",
+          "default": "ecmwf/hpc2020"
+        }
+      }      
     },
     "gribmodify": {
       "title": "Section for modify grib",
--- a/tactus/data/config_files/include/submission/atos_bologna.toml
+++ b/tactus/data/config_files/include/submission/atos_bologna.toml
@@ -87,6 +87,12 @@
   RTTOV_COEFDIR = "@RTTOV_COEFDIR@"
   UCX_LOG_LEVEL = "error"
 
+[submission.task_exceptions.IALBundleBuild.BATCH]
+  MEM = "#SBATCH --mem=0GB"
+  NODES = "#SBATCH --nodes=1"
+  NTASKS = "#SBATCH --ntasks=32"
+  WALLTIME = "#SBATCH --time=02:00:00"
+
 [submission.task_exceptions.InterpolSstSic]
   binary = "gl"
   bindir = "@BINDIR_GL@"
@@ -159,7 +165,7 @@
 [submission.types.parallel]
   NPROC = 32
   SCHOST = "hpc"
-  tasks = ["Forecast", "E927", "Pgd", "Prep", "C903"]
+  tasks = ["Forecast", "E927", "Pgd", "Prep", "C903", "IALBundleBuild"]
   WRAPPER = "srun"
 
 [submission.types.parallel.BATCH]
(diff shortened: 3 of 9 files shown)


trygveasp:
I am now using develop default, as it is supposed to work. Only compilation can be tested by:

```
tactus case ?tactus/data/config_files/configurations/cy49t2_arome tactus/data/config_files/modifications/compile.toml -o config.toml
tactus start suite --config-file config.toml
```

I'm now testing if it compiles as it has been claimed it should work.

There are some discussion topics on how we should be configuring the code compilation as it is connected to model cycle and platform and system specific settings.

At the moment there are:
```
[compile]
  arch = "ecmwf/hpc2020"
  git_branch = "develop"
  git_repo = "git@github.com:ACCORD-NWP/IAL"
  ial_dir = "@CASEDIR@/IAL"   # Default value, clones if not existing
```

git_branch is in conflict with the general setting CYCLE. However, there are no guaranties if they are consistent. So the CYCLE concept need some rethinking when we are in charge our own (possibly evolving) code.

The arch is system depending, but also possibly DP/SP etc

The ial_dir (local location of the clone)  is depending on the tactus system. A default value to `@CASEDIR@ ` is fine, but the question is how flexible it should be (I guess we should be able to use an existing clone) and how to make it consistent.

Then lastly and not implemented yet. The system installs in `@CASEDIR@/install` which should be the default location of binaries. But historically all binaries have been overridden in platform specific submission settings. This option should be kept, but maybe we need some restructuring? Comments and ideas @uandrae ?

uandrae:
Nice progress! Some comments and thoughts:

1. I guess one can use a tag or a hash for the `git_branch`? Perhaps use `version` as in bundle?
2. Can one derive the cycle from git and check the consistency with `general.cycle`?  If not we should couple `general.cycle` with settings in the compilation (try to) ensure consisteny.
3. We would have to agree if compilation should be default or not and configure bindir accordingly. If groups would like to have it done differently on local platforms they still have the freedom to do so.
4. I think we should always compile in both R32 and R64.
5.  `compile.arch` should probably be set per platform in some way. Can we couple the submission rules to this as well? Thinking about e.g. intel and gnu on atos.
6. How is compilation (de)activated?

trygveasp:
> Nice progress! Some comments and thoughts:
> 
>     1. I guess one can use a tag or a hash for the `git_branch`? Perhaps use `version` as in bundle?

At the moment it can be anything you can checkout
> 
>     2. Can one derive the cycle from git and check the consistency with `general.cycle`?  If not we should couple `general.cycle` with settings in the compilation (try to) ensure consisteny.
> 
In my opinon we should use something from the IAL code. Can version.txt be used?

>     3. We would have to agree if compilation should be default or not and configure bindir accordingly. If groups would like to have it done differently on local platforms they still have the freedom to do so.
> 
>     4. I think we should always compile in both R32 and R64.
> 
>     5. `compile.arch` should probably be set per platform in some way. Can we couple the submission rules to this as well? Thinking about e.g. intel and gnu on atos.
> 
A valid point. Now I guess it picks default, but this should be coordinated with ow ial_bundle is supposed to do this
>     6. How is compilation (de)activated?
```
[suite_control]
compile = false
```

uandrae:
`VERSIONS.txt` seems like the correct file. What's the procedure with this file @AlexandreMary? At the moment it says `50.0.0` which seems almost correct.

trygveasp:
Develop failed with:

```
/lus/h2resw01/scratch/sbu/tactus/CY49t2_AROME_nwp_DEMO_60x80_2500m_20250209/IAL/bundle/source/ial_dp/mpa/budgets/aro_suintbudget_omp.F90(49): error #7002: Error in opening
 the compiled module file.  Check INCLUDE paths.   [DDH_MIX]
USE DDH_MIX,ONLY: NPROMADDH, TYP_DDH
----^
/lus/h2resw01/scratch/sbu/tactus/CY49t2_AROME_nwp_DEMO_60x80_2500m_20250209/IAL/bundle/source/ial_dp/mpa/budgets/aro_suintbudget_omp.F90(56): error #6406: Conflicting attributes or multiple declaration of name.   [TYP_DDH]
TYPE(TYP_DDH),      INTENT(INOUT) :: YDDDH
-----^
/lus/h2resw01/scratch/sbu/tactus/CY49t2_AROME_nwp_DEMO_60x80_2500m_20250209/IAL/bundle/source/ial_dp/mpa/budgets/aro_suintbudget_omp.F90(76): error #6406: Conflicting attr
ibutes or multiple declaration of name.   [NPROMADDH]
NPROMADDH=KLON
...
```

trygveasp:
> Develop failed with:
> 
> ```
 But a re-run worked....

AlexandreMary:
> `VERSIONS.txt` seems like the correct file. What's the procedure with this file @AlexandreMary? At the moment it says `50.0.0` which seems almost correct.

Sorry, what's the question here ? This file is only used and updated by ECMWF until now, I believe

uandrae:
> > `VERSIONS.txt` seems like the correct file. What's the procedure with this file @AlexandreMary? At the moment it says `50.0.0` which seems almost correct.
> 
> Sorry, what's the question here ? This file is only used and updated by ECMWF until now, I believe

The question is how we from a git clone can determine which cycle we're on and VERSION.txt seems to be a suitable choice if we agree to keep it updated. If we stick to the current convention in the file I'd expected it to say `50.2.0` for the current develop.

AlexandreMary:
Well I guess it depends what you want to do with that, but the safest is probably to get this kind of info from git. That's what I implemented in IAL-build to make the link with gmkpack and find the latest root pack, for instance.

trygveasp:
> Well I guess it depends what you want to do with that, but the safest is probably to get this kind of info from git. That's what I implemented in IAL-build to make the link with gmkpack and find the latest root pack, for instance.

Can we really rely on getting this information from git? In git it can be called whatever. We also will run experiments based on user forks and branches and the script system needs to know which cycle this is based on.  
Can you point to where you do this in IAL-build?

AlexandreMary:
Again, it depends what you want to do from this information ?
50.0.0 seems to indicate that this is cycle 50, but what do you want to do from this ?
In IAL-build, it looks in the git history for the last tag that matches the legacy nomenclature eg. CY50T2 or CY50T2_something.nn, in order to look for a corresponding pack that matches.
Now if the user has added some tags that are not "official", that's his problem.

uandrae:
> Again, it depends what you want to do from this information ? 50.0.0 seems to indicate that this is cycle 50, but what do you want to do from this ? In IAL-build, it looks in the git history for the last tag that matches the legacy nomenclature eg. CY50T2 or CY50T2_something.nn, in order to look for a corresponding pack that matches. Now if the user has added some tags that are not "official", that's his problem.

We want to compare to the tactus `general.cycle` to see if the compiled code matches the expected cycle. Perhaps this IAL-build functionality could be used @trygveasp ?

trygveasp:
We could use the IAL-build way of getting latest cycle tag from history, or we could also get the ial-bundle create to create a file based on this information. In this way we will not have to implement the same logics.

Tactus has a legacy on DW where the cycle value defines the namelist input and the other expected input to the binary. This information is closely connected with the binaries. In DW we always use existing binaries, but in tactus we also want to create binaries from IAL source code. It's a bit similar to the davai settings, but as these are now inside the IAL source code they are hard-coded to the source code. We want these settings to be set picked up based on the cycle (like selecting the proper src/tasks/conf/atos_bologna.ini based  on cycle). However, I foresee some challenges and I do not agree that if a user changes something it's his/her problem. From a script point of view a system should be flexible enough.

There will be several CSCs running the code. UWC will likely have their own fork, and maybe also other naming schemes (e.g. today we have the h version). Should we always have an incremental system on top of the t version? That would be closer to the MF/vortex way and make collaboration easier.

AlexandreMary:
My point is: what happens in between releases, either on develop branch or on a user branch that introduces discrepancies ?
It's not anymore the latest cycle, and the VERSION.txt is not necessarily updated yet (with what ?). The only truthful source of information on "what version am I using" is Git. Whatever you are doing with this information.

trygveasp:
I fully see your point, but my point is even though git his the history we don't know, from a tactus point of view, which tags the repo maintainers have made, and we need to map IAL code against expected input to use tactus correct. The repo maintainers could be individual users or official CSCs and they could introduce conflicting tags.

Also we could have a situation where the IAL code is not a clone. Both ways of doing this only work if they are used as assumed.  I guess the only thing we can control from an ACCORD point of view is the IAL code in ACCORD-NWP repo and we could/should build on this.

Another tactus problem is that we don't know about what is used as general.cycle today until we have checked the IAL git history. One way of checking this might be to always set the cycle and do a consistency check on the cycle from git history when cloning/building. I still think it would be an idea that ial_bundle create made a file with what it from git history means the version is, which we could check against. Or a different script inside the bundle doing the same, in any case, I think the logic belongs to the bundle and not has to be duplicated.

uandrae:
Can we detach the cycle detection discussion from this particular PR?

uandrae:
Running the compilation in the standard suite fails with a trigger error.

uandrae:
I don't see any installation directory defined anywhere?

trygveasp:
> I don't see any installation directory defined anywhere?

At the moment it is set in:
https://github.com/trygveasp/tactus/blob/feature/ial_bundle_compilation/tactus/tasks/compilation.py#L71

If we compile we assume location of default binaries. But we need to decide on how the logic should be on how to override individual binaries, but I guess it could be done in e.g.:
https://github.com/trygveasp/tactus/blob/feature/ial_bundle_compilation/tactus/tasks/base.py#L175

trygveasp:
> Running the compilation in the standard suite fails with a trigger error.

What's the error? I have so far only tested compilation only by adding https://github.com/trygveasp/tactus/blob/feature/ial_bundle_compilation/tactus/data/config_files/modifications/compile.toml to tactus case command

uandrae:
Some erroneous trigger, you'll find out when you try ;-)

trygveasp:
> Some erroneous trigger, you'll find out when you try ;-)

tight lipped;-.) Fixed now in https://github.com/ACCORD-NWP/tactus/pull/11/commits/d1339a80a96c309ceb5fae770c92bba31fbe7bad

uandrae (approved):
LGTM!

trygveasp:
What do we do with e.g. changelog in this phase when we have a deode-workflow and a tactus script system?

uandrae:
We document what we do and merge update from DW. The question comes when we decide to tag something, what do we call it? I suggest that until we make the full separation on the DW we keep the current numbering but with some additional notation like `v0.26t.0` or similar. When tactus lives it's on life we can have a cleaner notation.

uandrae:
At least it shouldn't prevent us from do merges at this stage ;-)

trygveasp:
> At least it shouldn't prevent us from do merges at this stage ;-)

Well. the real question is should I just merge? Or should I update e.g. the changelog, and/or what else should I do?

uandrae:
Pleasae update the changelog.

PR #28, opened 2026-04-20
changes compilation task to use a tactus bundle that includes gl

pardallio:
## Describe your changes

Modifies compilation task to use a tactus specific bundle. This bundle file is provided in `tactus/data/compilation/@CYCLE@`

This removes the IALClone task as IAL is now downloaded as a part of the bundle but still allows for a local IAL folder to be given in the config file

## Type of change

### Testing

### Code quality

## Checklist for reviewers
- [ ] the code readable
- [ ] the code documented

  missing)

## Checklist for assignees
- [ ] the tests passing

## Tag possible reviewers


pardallio:
@uandrae here is the `gl-in-tactus-bundle` PR. It is still pointing to my gl branch but we can change that once that PR is merged

uandrae:
Can you review this @trygveasp ?

pardallio:
This will not work with the ACCORD IAL until PR #577 is merged

uandrae:
Testing this in https://github.com/uandrae/tactus/tree/tactus_minium gives a failure in build because `ecbundle` is not found. 

```
2026-04-29 11:18:16 | INFO     | Batch running  cd /scratch/snh/tactus/CY50t2_AROME_DEMO_60x80_2500m_20250209/bundle; ecbundle create  --bundle /etc/ecmwf/nfs/dh2_perm_b
/snh/git/tactus_accord/tactus/data/compilation/CY50t2/bundle.yml --update 
/bin/sh: ecbundle: command not found
```

pardallio:
I updated pyproject tom include ecbundle so you'll need to do poetry install or update

uandrae:
It's installed but it's not included in the path. If I force it to be the IAL code is still not cloned.

This is my case command using my `tactus_minimum` branch:
```
tactus case ?tactus/data/config_files/configurations/cy50t2_arome tactus/data/config_files/modifications/ww_modifs.toml  --start-suite --case-name test_build
```
Are you sure you don't happen to have ecbuild in your environment already? Second, the IAL code is not cloned.

pardallio:
when i run in the poetry shell environment 

`which ecbundle`

I get 

`/etc/ecmwf/nfs/dh2_home_b/prt2342/Documents/Dev/tactus/.venv/bin/ecbundle`

pardallio:
I just started a compilation from that branch  /home/prt2342/tactus_ecflow/jobout/test_build/Compilation/TactusBundleCreate.1

pardallio:
this commit is the culprit https://github.com/uandrae/tactus/commit/d7f97472d51d13ae677aaf957213af5485a32cea

that variable should be left empty, otherwise it will update the ial-source section on the bundle to the path given in ial_dir. This is meant for compilation of an existing clone of IAL

uandrae:
> this commit is the culprit [uandrae@d7f9747](https://github.com/uandrae/tactus/commit/d7f97472d51d13ae677aaf957213af5485a32cea)
> 
> that variable should be left empty, otherwise it will update the ial-source section on the bundle to the path given in ial_dir. This is meant for compilation of an existing clone of IAL

Ah, thanks, my bad!

uandrae:
> when i run in the poetry shell environment
> 
> `which ecbundle`
> 
> I get
> 
> `/etc/ecmwf/nfs/dh2_home_b/prt2342/Documents/Dev/tactus/.venv/bin/ecbundle`

So do I, but not from inside a batch job. For now I have a hack to get past this problem but I'm not convinced that e.g. /etc/ecmwf/nfs/dh2_home_b/prt2342/Documents/Dev/tactus/.venv/bin is automatically in the path inside a job.

trygveasp:
> > when i run in the poetry shell environment
> > `which ecbundle`
> > I get
> > `/etc/ecmwf/nfs/dh2_home_b/prt2342/Documents/Dev/tactus/.venv/bin/ecbundle`
> 
> So do I, but not from inside a batch job. For now I have a hack to get past this problem but I'm not convinced that e.g. /etc/ecmwf/nfs/dh2_home_b/prt2342/Documents/Dev/tactus/.venv/bin is automatically in the path inside a job.

No it's not. You call the task with that interpreter, so all python dependencies will come from the venv installation. But the environment is never loaded so that it enters path for a BatchJob object.

pardallio:
Ok, so  we either explicitly call ecbunlde with its full path using @TACTUS_HOME@ or export the venv path. Is there any reason why we wouldn not want to add venv to path or it is just because we did not need it so far?

uandrae:
> > > when i run in the poetry shell environment
> > > `which ecbundle`
> > > I get
> > > `/etc/ecmwf/nfs/dh2_home_b/prt2342/Documents/Dev/tactus/.venv/bin/ecbundle`
> > 
> > 
> > So do I, but not from inside a batch job. For now I have a hack to get past this problem but I'm not convinced that e.g. /etc/ecmwf/nfs/dh2_home_b/prt2342/Documents/Dev/tactus/.venv/bin is automatically in the path inside a job.
> 
> No it's not. You call the task with that interpreter, so all python dependencies will come from the venv installation. But the environment is never loaded so that it enters path for a BatchJob object.

OK, so we could implement a function that picks up this path and do e.g. 

```
[submission.task_exceptions.TactusBundleBuild]
  bindir = "resolve_venv_bin_path()"
[submission.task_exceptions.TactusBundleCreate]
  bindir = "resolve_venv_bin_path()"
```
And treat ecbundle with `get_binary()` inside compilation.py.

uandrae:
> Ok, so we either explicitly call ecbunlde with its full path using @TACTUS_HOME@ or export the venv path. Is there any reason why we wouldn not want to add venv to path or it is just because we did not need it so far?

That's the other option, to update `PATH`. Either is fine I think. Using `@TACTUS_HOME@` is not bullet proof. I often (always) put my .venv elsewhere.

trygveasp:
There is a third option as well and that is to make sure the path is set before the task is called (in the job). The subprocess in BatchJob inherits the environment it's called in.

pardallio:
It is also being called from within python, so there is surely a way to get the path to the python interpreter being used (which should be in the same location as ecbundle if im not mistaken)

trygveasp:
> It is also being called from within python, so there is surely a way to get the path to the python interpreter being used (which should be in the same location as ecbundle if im not mistaken)

Yes, either make sure the relative path to BatchJob is in path or set it from the outside. A possible problem is that in a plugin approach you might want more PATHS to be available

tactus/data/config_files/config.toml
```diff
@@ -114,6 +114,8 @@
   create_static_data = true
   create_time_dependent_suite = true
   do_archiving = true
+  compile = true
+  suite_definition = "CompilationSuiteDefinition"
```
trygveasp: Why have we defines this suite def in the general config file? Isn't this something to be added only when you want to do only compilation? compile = true is fine

tactus/data/compilation/CY50t2/bundle.yml
```diff
@@ -0,0 +1,242 @@
...
+            ENABLE_ECCODES_THREADS=ON
+            ECCODES_ENABLE_NETCDF=ON
+            ECCODES_ENABLE_FORTRAN=ON
+            ECCODES_ENABLE_MEMFS=ON
+            ECCODES_INSTALL_EXTRA_TOOLS=ON
+            ENABLE_AEC=OFF
+            ECCODES_ENABLE_JPG_LIBOPENJPEG=OFF
+
+    - ial-source :
+        git     : ${GITHUB}/pardallio/ACCORD-IAL
```
trygveasp: I like the idea that this is taken from CVS, but I do foresee that we would like to test with a local IAL code also existing. Is it possible with this setup?

trygveasp: I now see that you still have ial_dir in the bundle creation, but I am not able to really understand how it works yet:-)

tactus/tasks/compilation.py
```diff
@@ -19,56 +21,64 @@ def __init__(self, config):
...
-            batch_job = BatchJob(os.environ)
-            cmd = f"git clone {self.git_ial_repo} {self.ial_dir}"
-            cmd = cmd.replace("[TOKEN]", self.git_token)
-            batch_job.run(cmd)
-            batch_job.run(f"cd {self.ial_dir}; git checkout {self.git_ial_branch}")
+        bundle_file = self.config["compile.bundle_file"]
+        bundle_file = self.platform.substitute(bundle_file)
 
+        if self.config["compile.ial_dir"]:
+            with open(bundle_file, "r") as f:
```
trygveasp: need encoding

tactus/tasks/compilation.py
```diff
@@ -19,56 +21,64 @@ def __init__(self, config):
...
-        if git_token:
-            git_token_str = f"--github-token {git_token}"
-        self.git_token_str = git_token_str
-        self.ial_dir = self.platform.substitute(ial_dir)
+                        # Add the new key
+                        ial["dir"] = self.platform.substitute(
+                            self.config["compile.ial_dir"]
+                        )
+
+            with open(bundle_file, "w") as f:
```
trygveasp: need encoding

tactus/data/compilation/CY50t2/bundle.yml
pardallio: The default bundle file defines a git source for IAL. 

if you want to do a local copy, you can either change the bundle file, or do it via the tactus config file with the ial_dir parameter from the compile section

tactus/data/config_files/config.toml
pardallio: Yes it is, I forgot to remove it.

tactus/tasks/compilation.py
pardallio: I'm not sure what you mean by this comment. How do we encode?

trygveasp: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/unspecified-encoding.html

trygveasp:
I'm testing this on my laptop to also get gl built, and I have a request and a problem. Could we make --ninja optional? 

And when using my local IAL code (which compiles and works so far), it does not work with using an existing ial_dir since I have architecture files in my bundle directory I need to use them. But instead a symlink is created to an non-existing arch directory in the experiment. So  I need to copy/symlink them in manually. Shouldn't it respect the bundle directory found in ial_dir?

trygveasp:
gl fails since it assumes intel compiler options. I need to tell it that I use gfortran. Seems to be no tests on this in CMakeLists.txt in gl?

pardallio:
I agree to leave ninja as optional. I also prefer regular cmake.

ecbundle expects an arch directory to be palces next to the bundle file. I have created an issue to pass arch dir as an argument to ecbundle create but the feature has not been implemented yet. https://github.com/ecmwf/ecbundle/issues/36.

Nevertheless, it also accepts relative paths with regard to the folder where `bundle build`  is called. So,
it should pick up the arch directory in your local IAL because we set arch-dir for `bundle build` as `source/ial-source/bundle/arch/ecmwf/hpc2020` can you share the log file?

trygveasp:
> gl fails since it assumes intel compiler options. I need to tell it that I use gfortran. Seems to be no tests on this in CMakeLists.txt in gl?

gl compiled with gfortran by hacking in `gl/CMakeLists.txt`:
```

#target_compile_options(gl.${PREC} PUBLIC
#    "SHELL:-assume realloc_lhs,byterecl"
#    -fpic 
#    "SHELL:-align array64byte,all"
#    -qopenmp 
#    "SHELL:-qopenmp-threadprivate compat"
#    "SHELL:-fp-model source" -fpp
#    -qopt-report=5 -qopt-report-phase=vec
#    -ftz -diag-disable=remark,cpu-dispatch)

target_compile_options(gl.${PREC} PUBLIC
    -fpic
    -fopenmp
    -cpp
    -fallow-argument-mismatch
)
```

pardallio:
Yes, i only implemented bundled gl compilation for intel

pardallio:
I added the check for the compiler in my gl branch in the meantime

pardallio:
@uandrae @trygveasp , I have one question. Right now compilation is happening in a set directory in scratch by default (`@CASEDIR@/bundle`). 

with the build files going to `@CASEDIR@/bundle/build` and the installation to `@CASEDIR@/bundle/install`. This is very useful to re-start a failed compilation but can lead to problems if we do the same case months apart and IAL has some changes that require the whole build folder to be removed. I think this can be fixed if we remove the build directory after a successfull installation maybe.

Also, it seems to me that by placing the installation inside CASEDIR can make us compile the  IAL version multiple times if we run multiple cases. Shouldn't we have the installation in some sort of cache directory so that multiple cases can use the same binary?

If we do that the we would need to decide how to properly name the installation folder. The IAL commit hash could be a valid option, but with the bundle we have multiple combinations of versions of subprojects, so I don't really know the solution for this.

trygveasp:
> 
> with the build files going to `@CASEDIR@/bundle/build` and the installation to `@CASEDIR@/bundle/install`. This is very useful to re-start a failed compilation but can lead to problems if we do the same case months apart and IAL has some changes that require the whole build folder to be removed. I think this can be fixed if we remove the build directory after a successfull installation maybe.
>

I think the user should be able to remove if needed or we should make a force reinstall option. The use case here is probably more that the user will do local changes or update repositories and not re-compile everything
 
> Also, it seems to me that by placing the installation inside CASEDIR can make us compile the IAL version multiple times if we run multiple cases. Shouldn't we have the installation in some sort of cache directory so that multiple cases can use the same binary?
> 

We already have that. The binaries can be overridden. Just set them and skip compilation

> If we do that the we would need to decide how to properly name the installation folder. The IAL commit hash could be a valid option, but with the bundle we have multiple combinations of versions of subprojects, so I don't really know the solution for this.

Do we need that? What we do need is handling of DP/SP installations.

pardallio:
> We already have that. The binaries can be overridden. Just set them and skip compilation

True, but it is cumbersome to set a custom install_dir that points to some other casedir everytime you start a new case. Coulnd't it be made automatically?  We already have something like that for IFS data to avoid doing marsrep all over again

trygveasp:
> > We already have that. The binaries can be overridden. Just set them and skip compilation
> 
> True, but it is cumbersome to set a custom install_dir that points to some other casedir everytime you start a new case. Coulnd't it be made automatically? We already have something like that for IFS data to avoid doing marsrep all over again

If you set the general binary path in your submission file to your reference binaries path they will be used when you do not compile in the experiment. Will that not solve your problem and be relatively automatic?

pardallio:
no,  because I don't necessarily have a reference binary path. 

The same way I don't necessarily have an IFS data reference path, so tactus creates one for me and puts there the data for each date that I run, and I don't need to update any config file manually to avoid having to download that all over again.

pardallio:
I've just commited a possible implementation.

It checks the bundle and gets the valid commit hash for each repo. then merges all of them into a unique hash. If there are changes in any repo it flags the hash as dirty

then it does the build in `/scratch/prt2342/tactus/bundle_cache/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/81bc8f3658d998ad20bf3a42d3a12f20ee3759e9b5529acac38298f2926dd942/build`

trygveasp:
> no, because I don't necessarily have a reference binary path.
> 
> The same way I don't necessarily have an IFS data reference path, so tactus creates one for me and puts there the data for each date that I run, and I don't need to update any config file manually to avoid having to download that all over again.

I think we are misunderstanding each other here. Maybe we should have a chat next week. 

You wanted to reuse existing binaries so you do have "reference" binaries. If you see in e.g. https://github.com/pardallio/tactus/blob/tactus-bundle/tactus/tasks/base.py#L245 the submission.bindir is the last one used. So by setting that to the installation directory, which you want to re-use binaries from, you will get them automatically (as long as you don't ask tactus to compile new ones in your experiment).

pardallio:
Yes, I understand.

It is possible to set the general binary path to whatever folder we have manually, but what I am suggesting is caching the builds from one experiment to be able to use them on other experiments and having tactus find these builds automatically.

we can discuss better in the meeting maybe

pardallio:
right now it's working mostly as intended. but still need to deal with building single/double precision binaries

tactus/tasks/compilation.py
```diff
@@ -105,32 +102,33 @@ def __init__(self, config):
...
             default_link = arch_dir / "default"
             if default_link.exists() and default_link.is_symlink():
-                arch=str(default_link.resolve())
+                arch = str(default_link.resolve())
             else:
-                arch=str(arch_dir)
+                arch = str(arch_dir)
             arch = arch.split("arch")[-1]
-            compile_dir=f"{self.config['compile.cache_dir']}/{arch}/{self.bundle_hash}"
+            compile_dir = f"{self.config['compile.cache_dir']}/{arch}/{self.bundle_hash}"
```
trygveasp, line 196: I am not sure if the last element of the arch path is what we always want?

trygveasp:
I think I now understand more what you mean by your cache and the hash creation. How I see the building of the tactus bundle is that the user is fully in charge of the local compilation. It means if the user want to add something, the user will recreate/rebuild the bundle and the experiment is testing one test case. If same binaries should be used in a different case, mechanisms exist for this.  

I see the need for a forced recreation of the bundle and/or the build directory, but in my option the hashes will just cache up old builds which are probably not needed. I might still have misunderstood something though:-)

tactus/tasks/compilation.py
pardallio: it's the last element of the split. contains all the path in front of "arch" or "extra-arch" which uniquely identifies  the build.

pardallio:
The cached build are stored in scratch which is supposed to be erased periodically, so old builds will not build up.

To me, compilation is a 30min task which can most of the time be avoided, and it seems more correct to have it in some reference folder instead of being linked to a specific case.

tactus/tasks/compilation.py
trygveasp: So from `arch/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0/ ` you want to identify it as 2.9.0 ?

The arch can be specified in several ways as I understand it.

pardallio: in that case it will evaluate to `ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0` and the build will end up in`$SCRATCH/tactus/bundle_cache/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0/81bc8f3658d998ad20bf3a42d3a12f20ee3759e9b5529acac38298f2926dd942/build`

trygveasp:
> 
> To me, compilation is a 30min task which can most of the time be avoided, and it seems more correct to have it in some reference folder instead of being linked to a specific case.

If the repo is dirty you will then not recompile although you might have local changes? At least this need to work or else the user will be surprised

As I see this it is a matter of trusting the user or not. If the user does not need recompilation he/she can turn it off (and skip the tasks and save time). The implementation here allows to always have it on, but do the task which will be faster if nothing needs to be done. I will not be against it, but I also do not see it as strictly necessary.

tactus/tasks/compilation.py
trygveasp: Sorry, that was my bad. Split splits default on space and for some reason I thought you were splitting on slashes... Then my question is, why the last element? Do we have a list of architectures now?

pardallio:
> > To me, compilation is a 30min task which can most of the time be avoided, and it seems more correct to have it in some reference folder instead of being linked to a specific case.
> 
> If the repo is dirty you will then not recompile although you might have local changes? At least this need to work or else the user will be surprised
> 
> As I see this it is a matter of trusting the user or not. If the user does not need recompilation he/she can turn it off (and skip the tasks and save time). The implementation here allows to always have it on, but do the task which will be faster if nothing needs to be done. I will not be against it, but I also do not see it as strictly necessary.

You are correct, if the repo is already marked as dirty, it will not recompile unless forced to do so. I did not think of that case.
I made it this way because cmake takes about 2 minutes to check that everything is built already, and by skipping this we reduced the task duration from 4 to 2 minutes in case of an already completed build. 

 It is better then to do the opposite: add the option to skip compilation, and make it default to false.

trygveasp:
> 
> It is better then to do the opposite: add the option to skip compilation, and make it default to false.

That option is already there:
```
[suite_control]
compile = false
```

The default value should be based on what we think is the normal procedure. To start from existing binaries or build our own. If binaries are built we could let the user decide if he/she wants to set it False before restarting. A cache (as implemented here) is also an alternative but we need to be able to pick up local changes for recompilation.

trygveasp:
I tried it on ATOS-Bologna and got this:

```
2026-05-12 13:23:49 | INFO     | Batch running srun /etc/ecmwf/nfs/dh1_perm_a/deployde330/github-actions/install/CY50t2/intel/R32/refs/pull/38/merge/bin/MASTERODB
/etc/ecmwf/nfs/dh1_perm_a/deployde330/github-actions/install/CY50t2/intel/R32/refs/pull/38/merge/bin/MASTERODB: error while loading shared libraries: libecflow_lightf.so: cannot open shared object file: No such file or directory
```

So it is not using the installed binary by me

pardallio:
Is that for the Forecast task?

The config file still points to the deode installation on that case

trygveasp:
> Is that for the Forecast task?
> 
> The config file still points to the deode installation on that case

Yes it is. I see in https://github.com/pardallio/tactus/blob/tactus-bundle/tactus/data/config_files/include/submission/atos_bologna.toml#L76 that it should use that binary. Has to be removed IMO, but is maybe already handled @uandrae  ?

uandrae:
I agree for CY50, but for CY49T2 I think we should keep the references to the deode binaries for convenience for now. Not sure how to implement this in a neat way though. Seems like we need a `@CYCLE@_@HOST@.toml` file for this.

In addition I think we should compile CY50 in both SP/DP but I guess that's in the plan?

trygveasp:
> I agree for CY50, but for CY49T2 I think we should keep the references to the deode binaries for convenience for now. Not sure how to implement this in a neat way though. Seems like we need a `@CYCLE@_@HOST@.toml` file for this.
> 
> In addition I think we should compile CY50 in both SP/DP but I guess that's in the plan?

I tested cy50t2. Only changed the domain:
`tactus case ?tactus/data/config_files/configurations/cy50t2_harmonie_arome TEST_40_50.toml -o config.toml --cace-name pr28 --start-suite`

uandrae:
As this PR sets compile to default I'd suggest to put `"@CASEDIR@/install/bin"` in the json schema for bindir ( and bindir_gl if this is still to be used ). For the CY49T2 needs I think we could add the following to `tactus/data/config_files/modifications/cycle/CY49t2.toml`.

```
[submission]
  bindir = "@INSTALL_DIR@/@CYCLE@/@COMPILER@/R64/@IAL_VERSION@/bin"
  bindir_gl = "@INSTALL_DIR@/gl/@COMPILER@/@GL_VERSION@/bin"

[submission.task_exceptions.Forecast]
  bindir = "@INSTALL_DIR@/@CYCLE@/@COMPILER@/@PRECISION@/@IAL_VERSION@/bin"
```
It's deode specific but we have no plans to work on CY49T2 here anyway.

If you agree you could perhaps add this @pardallio. Including cleaning tactus/data/config_files/include/submission/atos_bologna.toml`.

pardallio:
Yes, I agree to change that to change the case install dir to default.

pardallio:
as for the precision, maybe compile both only when single precision is asked (whihc is the default)

uandrae (commented):
Very nice to have a clean build with also gl, and other tools, included!

The usefulness of the cache depends very much on how one works and the lifetime of data on the disks. How would it work on atos if a cache has been cleaned? Would it just work of would it fail due to missing files or similar? The cache could be case specific by a simple config change. I'd suggest that we try this out and see what works in practice. Some general comments:

- It would be good with some documentation on the compilation aspects. 
- I think it would be useful copy the bundle.yaml to the corresponding hashed directory. At some point there might be a lot of directories and they will be difficult to backtrack.
- Now we have a duplicated structure with both R64/R32 and _dp/_sp suffixes. Can we get rid of the suffixes?

uandrae:
I introduced the following change to test the bundle cache

```
[snh@ac6-102 tactus_pr28]$ git diff 
diff --git a/tactus/data/compilation/CY50t2/bundle.yml b/tactus/data/compilation/CY50t2/bundle.yml
index a9d44fc4..904897ab 100644
--- a/tactus/data/compilation/CY50t2/bundle.yml
+++ b/tactus/data/compilation/CY50t2/bundle.yml
@@ -53,7 +53,7 @@ projects :
             ECCODES_ENABLE_JPG_LIBOPENJPEG=OFF
 
     - ial-source :
-        git     : ${GITHUB}/pardallio/ACCORD-IAL
+        git     : ${GITHUB}/uandrae/ACCORD-IAL
         version : feature/toolchain-flags
         bundle  : false
         gmkpack : 'src/local'
```
Apart from the fact that I don't have a repo name `ACCORD-IAL` I expected this to be a valid change. However it fails with the following error

```
ESC[95mCheckout project github.com:/uandrae/ACCORD-IAL @ feature/toolchain-flagsESC[0m
ESC[93mWARNING: ed653566da7d4505ab09e12b947ceb5d is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
ESC[94m+ git remote add ed653566da7d4505ab09e12b947ceb5d git@github.com:/uandrae/ACCORD-IALESC[0m
ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
ESC[91mERROR: Could not download or update ial-source ...ESC[0m
```
After correcting the remote  to `uandrae/IAL`, but keeping the branch, I still get the same error:

```
ESC[95mCheckout project github.com:/uandrae/IAL @ feature/toolchain-flagsESC[0m
ESC[93mWARNING: eeec494cba20d4c7ae560cc38b7a8b14 is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
ESC[94m+ git remote add eeec494cba20d4c7ae560cc38b7a8b14 git@github.com:/uandrae/IALESC[0m
ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
ESC[91mERROR: Could not download or update ial-source ...ESC[0m
```
What is the expected user behaviour here?

uandrae:
I tried using latest develop@IAL instead of your branch in your repo but it failed in the bundle.

hpc-login:/lus/h2resw01/scratch/snh/tactus/bundle_cache/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0/669bf2552ba7ba080c6df9e2164b643fe8af42d348be8ebc0947ea5ff9a056f2/build/R64/CMakeFiles/CMakeError.log

pardallio:
> Very nice to have a clean build with also gl, and other tools, included!
> 
> The usefulness of the cache depends very much on how one works and the lifetime of data on the disks. How would it work on atos if a cache has been cleaned? Would it just work of would it fail due to missing files or similar? The cache could be case specific by a simple config change. I'd suggest that we try this out and see what works in practice. Some general comments:
> 
>     * It would be good with some documentation on the compilation aspects.
> 
>     * I think it would be useful copy the bundle.yaml to the corresponding hashed directory. At some point there might be a lot of directories and they will be difficult to backtrack.
> 
>     * Now we have a duplicated structure with both R64/R32 and _dp/_sp suffixes. Can we get rid of the suffixes?

The suffixes are set within the multiple repos of the subprojects. That change would require coordination between IAL/SURFEX/PHYEX/FALFILFA. It can be done but may take a while

pardallio:
> I introduced the following change to test the bundle cache
> 
> ```
> [snh@ac6-102 tactus_pr28]$ git diff 
> diff --git a/tactus/data/compilation/CY50t2/bundle.yml b/tactus/data/compilation/CY50t2/bundle.yml
> index a9d44fc4..904897ab 100644
> --- a/tactus/data/compilation/CY50t2/bundle.yml
> +++ b/tactus/data/compilation/CY50t2/bundle.yml
> @@ -53,7 +53,7 @@ projects :
>              ECCODES_ENABLE_JPG_LIBOPENJPEG=OFF
>  
>      - ial-source :
> -        git     : ${GITHUB}/pardallio/ACCORD-IAL
> +        git     : ${GITHUB}/uandrae/ACCORD-IAL
>          version : feature/toolchain-flags
>          bundle  : false
>          gmkpack : 'src/local'
> ```
> 
> Apart from the fact that I don't have a repo name `ACCORD-IAL` I expected this to be a valid change. However it fails with the following error
> 
> ```
> ESC[95mCheckout project github.com:/uandrae/ACCORD-IAL @ feature/toolchain-flagsESC[0m
> ESC[93mWARNING: ed653566da7d4505ab09e12b947ceb5d is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
> ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
> ESC[94m+ git remote add ed653566da7d4505ab09e12b947ceb5d git@github.com:/uandrae/ACCORD-IALESC[0m
> ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
> ESC[91mERROR: Could not download or update ial-source ...ESC[0m
> ```
> 
> After correcting the remote to `uandrae/IAL`, but keeping the branch, I still get the same error:
> 
> ```
> ESC[95mCheckout project github.com:/uandrae/IAL @ feature/toolchain-flagsESC[0m
> ESC[93mWARNING: eeec494cba20d4c7ae560cc38b7a8b14 is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
> ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
> ESC[94m+ git remote add eeec494cba20d4c7ae560cc38b7a8b14 git@github.com:/uandrae/IALESC[0m
> ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
> ESC[91mERROR: Could not download or update ial-source ...ESC[0m
> ```
> 
> What is the expected user behaviour here?

That is an issue with ecbundle, not the bundle cache. The bundle cache is only applied to the build and install folders.
For the bundle itself it is more tricky to build a hash before cloning the bundle sources.
There is surely a flag on ecbundle that avoids that issue

uandrae:
> > Very nice to have a clean build with also gl, and other tools, included!
> > The usefulness of the cache depends very much on how one works and the lifetime of data on the disks. How would it work on atos if a cache has been cleaned? Would it just work of would it fail due to missing files or similar? The cache could be case specific by a simple config change. I'd suggest that we try this out and see what works in practice. Some general comments:
> > ```
> > * It would be good with some documentation on the compilation aspects.
> > 
> > * I think it would be useful copy the bundle.yaml to the corresponding hashed directory. At some point there might be a lot of directories and they will be difficult to backtrack.
> > 
> > * Now we have a duplicated structure with both R64/R32 and _dp/_sp suffixes. Can we get rid of the suffixes?
> > ```
> 
> The suffixes are set within the multiple repos of the subprojects. That change would require coordination between IAL/SURFEX/PHYEX/FALFILFA. It can be done but may take a while

Perhaps not a large problem for now as we're only using R32 version of MASTERODB but one option would be a renaming on our side post build.

uandrae:
> > I introduced the following change to test the bundle cache
> > ```
> > [snh@ac6-102 tactus_pr28]$ git diff 
> > diff --git a/tactus/data/compilation/CY50t2/bundle.yml b/tactus/data/compilation/CY50t2/bundle.yml
> > index a9d44fc4..904897ab 100644
> > --- a/tactus/data/compilation/CY50t2/bundle.yml
> > +++ b/tactus/data/compilation/CY50t2/bundle.yml
> > @@ -53,7 +53,7 @@ projects :
> >              ECCODES_ENABLE_JPG_LIBOPENJPEG=OFF
> >  
> >      - ial-source :
> > -        git     : ${GITHUB}/pardallio/ACCORD-IAL
> > +        git     : ${GITHUB}/uandrae/ACCORD-IAL
> >          version : feature/toolchain-flags
> >          bundle  : false
> >          gmkpack : 'src/local'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Apart from the fact that I don't have a repo name `ACCORD-IAL` I expected this to be a valid change. However it fails with the following error
> > ```
> > ESC[95mCheckout project github.com:/uandrae/ACCORD-IAL @ feature/toolchain-flagsESC[0m
> > ESC[93mWARNING: ed653566da7d4505ab09e12b947ceb5d is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
> > ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
> > ESC[94m+ git remote add ed653566da7d4505ab09e12b947ceb5d git@github.com:/uandrae/ACCORD-IALESC[0m
> > ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
> > ESC[91mERROR: Could not download or update ial-source ...ESC[0m
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After correcting the remote to `uandrae/IAL`, but keeping the branch, I still get the same error:
> > ```
> > ESC[95mCheckout project github.com:/uandrae/IAL @ feature/toolchain-flagsESC[0m
> > ESC[93mWARNING: eeec494cba20d4c7ae560cc38b7a8b14 is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
> > ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
> > ESC[94m+ git remote add eeec494cba20d4c7ae560cc38b7a8b14 git@github.com:/uandrae/IALESC[0m
> > ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
> > ESC[91mERROR: Could not download or update ial-source ...ESC[0m
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What is the expected user behaviour here?
> 
> That is an issue with ecbundle, not the bundle cache. The bundle cache is only applied to the build and install folders. For the bundle itself it is more tricky to build a hash before cloning the bundle sources. There is surely a flag on ecbundle that avoids that issue

OK, but it's still a usage related problem. One needs some guidance what's allowed to change without cleaning the working directories.

pardallio:
> > > I introduced the following change to test the bundle cache
> > > ```
> > > [snh@ac6-102 tactus_pr28]$ git diff 
> > > diff --git a/tactus/data/compilation/CY50t2/bundle.yml b/tactus/data/compilation/CY50t2/bundle.yml
> > > index a9d44fc4..904897ab 100644
> > > --- a/tactus/data/compilation/CY50t2/bundle.yml
> > > +++ b/tactus/data/compilation/CY50t2/bundle.yml
> > > @@ -53,7 +53,7 @@ projects :
> > >              ECCODES_ENABLE_JPG_LIBOPENJPEG=OFF
> > >  
> > >      - ial-source :
> > > -        git     : ${GITHUB}/pardallio/ACCORD-IAL
> > > +        git     : ${GITHUB}/uandrae/ACCORD-IAL
> > >          version : feature/toolchain-flags
> > >          bundle  : false
> > >          gmkpack : 'src/local'
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Apart from the fact that I don't have a repo name `ACCORD-IAL` I expected this to be a valid change. However it fails with the following error
> > > ```
> > > ESC[95mCheckout project github.com:/uandrae/ACCORD-IAL @ feature/toolchain-flagsESC[0m
> > > ESC[93mWARNING: ed653566da7d4505ab09e12b947ceb5d is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
> > > ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
> > > ESC[94m+ git remote add ed653566da7d4505ab09e12b947ceb5d git@github.com:/uandrae/ACCORD-IALESC[0m
> > > ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
> > > ESC[91mERROR: Could not download or update ial-source ...ESC[0m
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > After correcting the remote to `uandrae/IAL`, but keeping the branch, I still get the same error:
> > > ```
> > > ESC[95mCheckout project github.com:/uandrae/IAL @ feature/toolchain-flagsESC[0m
> > > ESC[93mWARNING: eeec494cba20d4c7ae560cc38b7a8b14 is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
> > > ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
> > > ESC[94m+ git remote add eeec494cba20d4c7ae560cc38b7a8b14 git@github.com:/uandrae/IALESC[0m
> > > ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
> > > ESC[91mERROR: Could not download or update ial-source ...ESC[0m
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > What is the expected user behaviour here?
> > 
> > 
> > That is an issue with ecbundle, not the bundle cache. The bundle cache is only applied to the build and install folders. For the bundle itself it is more tricky to build a hash before cloning the bundle sources. There is surely a flag on ecbundle that avoids that issue
> 
> OK, but it's still a usage related problem. One needs some guidance what's allowed to change without cleaning the working directories.

Yes, the ecbundle command that tactus runs should probably have that flag if it exists, or have the option to enable it via config. I'm looking into it at the moment

pardallio:
> I tried using latest develop@IAL instead of your branch in your repo but it failed in the bundle.
> 
> hpc-login:/lus/h2resw01/scratch/snh/tactus/bundle_cache/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0/669bf2552ba7ba080c6df9e2164b643fe8af42d348be8ebc0947ea5ff9a056f2/build/R64/CMakeFiles/CMakeError.log

I haven't tried to compile with the latest develop yet. but i will look into it

pardallio:
> > > > I introduced the following change to test the bundle cache
> > > > ```
> > > > [snh@ac6-102 tactus_pr28]$ git diff 
> > > > diff --git a/tactus/data/compilation/CY50t2/bundle.yml b/tactus/data/compilation/CY50t2/bundle.yml
> > > > index a9d44fc4..904897ab 100644
> > > > --- a/tactus/data/compilation/CY50t2/bundle.yml
> > > > +++ b/tactus/data/compilation/CY50t2/bundle.yml
> > > > @@ -53,7 +53,7 @@ projects :
> > > >              ECCODES_ENABLE_JPG_LIBOPENJPEG=OFF
> > > >  
> > > >      - ial-source :
> > > > -        git     : ${GITHUB}/pardallio/ACCORD-IAL
> > > > +        git     : ${GITHUB}/uandrae/ACCORD-IAL
> > > >          version : feature/toolchain-flags
> > > >          bundle  : false
> > > >          gmkpack : 'src/local'
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Apart from the fact that I don't have a repo name `ACCORD-IAL` I expected this to be a valid change. However it fails with the following error
> > > > ```
> > > > ESC[95mCheckout project github.com:/uandrae/ACCORD-IAL @ feature/toolchain-flagsESC[0m
> > > > ESC[93mWARNING: ed653566da7d4505ab09e12b947ceb5d is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
> > > > ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
> > > > ESC[94m+ git remote add ed653566da7d4505ab09e12b947ceb5d git@github.com:/uandrae/ACCORD-IALESC[0m
> > > > ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
> > > > ESC[91mERROR: Could not download or update ial-source ...ESC[0m
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > After correcting the remote to `uandrae/IAL`, but keeping the branch, I still get the same error:
> > > > ```
> > > > ESC[95mCheckout project github.com:/uandrae/IAL @ feature/toolchain-flagsESC[0m
> > > > ESC[93mWARNING: eeec494cba20d4c7ae560cc38b7a8b14 is a new remote for repository ial-source. Be careful what you wish for! (check following lines)ESC[0m
> > > > ESC[94m+ cd /lus/h2resw01/scratch/snh/tactus/pr28/bundle/source/ial-sourceESC[0m
> > > > ESC[94m+ git remote add eeec494cba20d4c7ae560cc38b7a8b14 git@github.com:/uandrae/IALESC[0m
> > > > ESC[91mERROR: Branch feature/toolchain-flags was already tracking origin/feature/toolchain-flags. Manual intervention needed.ESC[0m
> > > > ESC[91mERROR: Could not download or update ial-source ...ESC[0m
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > What is the expected user behaviour here?
> > > 
> > > 
> > > That is an issue with ecbundle, not the bundle cache. The bundle cache is only applied to the build and install folders. For the bundle itself it is more tricky to build a hash before cloning the bundle sources. There is surely a flag on ecbundle that avoids that issue
> > 
> > 
> > OK, but it's still a usage related problem. One needs some guidance what's allowed to change without cleaning the working directories.
> 
> Yes, the ecbundle command that tactus runs should probably have that flag if it exists, or have the option to enable it via config. I'm looking into it at the moment

It turns out the ecbundle flag I had in my did not solve the issue. It seems that ecbundle cannot handle it when we change the remote, but keep the branch name. 
I'm not sure how to proceed here. It seems to be a limitation on ecbundle but we could overcome out by removing the bundle source folder with a flag in the config file.
On the other hand it states in the error message that manual intervention is needed, so we could also just add this as a warning to the documentation.

dhaumont:
> > I tried using latest develop@IAL instead of your branch in your repo but it failed in the bundle.
> > hpc-login:/lus/h2resw01/scratch/snh/tactus/bundle_cache/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0/669bf2552ba7ba080c6df9e2164b643fe8af42d348be8ebc0947ea5ff9a056f2/build/R64/CMakeFiles/CMakeError.log
> 
> I haven't tried to compile with the latest develop yet. but i will look into it

I'm on it

dhaumont (changes requested):
I think some unit tests are missing. You can mock the compilation step to make it faster, but unit tests are important to ensure that the compilation pipeline is working as expected.

tactus/data/compilation/CY50t2/bundle.yml
```diff
@@ -0,0 +1,236 @@
+---
+### Bundle
+# Variable $GITHUB shall be set by the user according to the protocol to be used
+# to connect to Github, e.g. GITHUB="https://github.com" to use https and tokens.
+
+name    : ial-bundle
```
dhaumont: I'm a bit concerned with the proliferation of bundle.yml files. In the future, it will be very complicated to know which one have been used to compile a given version of MASTERODB. We decided to move the bundle.yml from IAL-bundle repository into IAL/bundle, to have IAL self contained. It was a good move in my opinion, but here you extract bundle.yml again to an outside repository.
What is the purpose of doing this ? Can't we use the IAL/bundle/bundle.yml, to keep things simple ?
@uandrae @pardallio

pardallio: I agree, but the dial-bundle  does not have gl.

dhaumont: For me, IAL and tactus should go hand in hand: if something is needed to run tactus, than IAL/bundle/bundle.yml should contain it.
 Why don't you simply add gl into IAL/bundle/bundle.yml, but activate it with a new additional option `with_gl`  at the end of  bundle.yml  ?

pardallio: Since gl is not needed to run IAL, it was decided that the bundle in IAL should not contain it.

dhaumont: If it forces us to duplicate the bundle.yml, I'm afraid this decision should be reconsidered...Or maybe we should create a new bundle for the GL related stuff ?

uandrae: One reason is because I asked if we could include gl in the compilation, and I don't think we should not include gl in IAL.

dhaumont: As of today, the local bundle from this PR and the one from IAL/develop are very different, and it is hard to tell if it is on purpose or not. Can you update the bundle.yml to align it with the one from IAL/develop  ?

dhaumont: > One reason is because I asked if we could include gl in the compilation, and I don't think we should not include gl in IAL.

Not sure to understand your sentence Ulf, it contains too many double negations for me to handle :-)
What do you mean ?

uandrae: I'm fine with treating gl separately if it allows consistency in the IAL part.

uandrae: > > One reason is because I asked if we could include gl in the compilation, and I don't think we should not include gl in IAL.
> 
> Not sure to understand your sentence Ulf, it contains too many double negations for me to handle :-) What do you mean ?

Sorry: We should not include gl in IAL.

pardallio: There are more differences other than just adding gl

It is also different because tactus will require a different eccodes version, for example. 

It allows us to have a bundle specifically for tactus and its needs.

dhaumont: Tactus and IAL are now the two main tools composing the ACCORD-NWP forecast suite, and are the two faces of the same coin. I believe they should be aligned. Why would you use different versions of eccodes between IAL and tactus ? Maybe I'm missing something and we should just continue the discussion in a meeting.

uandrae:
For reference I've managed to complete a AROME forecast with develop after removing some fullpos variables, #50, and some binary ( lfitools_dp) fiddling. For the latter part I think we should let this PR take care of the updates but it's a nice step forward!

pardallio:
This PR is now ready for re-review

dhaumont:
@pardallio The branch to compile is specified at two different places:
1. https://github.com/ACCORD-NWP/tactus/blob/cf20b5ee441bbabdca5739842fe87b6f67d2b64e/tactus/data/config_files/config.toml#L56
2. https://github.com/ACCORD-NWP/tactus/blob/cf20b5ee441bbabdca5739842fe87b6f67d2b64e/tactus/data/config_files/modifications/compile.toml#L2

The 2. is redundant and should be removed

pardallio:
can anyone review?

uandrae (approved):
LGTM! Please update gl and changelog before merge.

dhaumont (approved):
LGTM

PR #75, opened 2026-05-27
Feature/add pixi

mfroelund:
## Describe your changes
This PR introduced the possibility to manage the tactus project and its dependencies using `pixi` (https://pixi.prefix.dev/latest/). 

## Type of change

### Testing

### Code quality

## Checklist for reviewers
- [ ] the code readable
- [ ] the code documented

  missing)

## Checklist for assignees
- [ ] the tests passing

## Tag possible reviewers


mfroelund:
Before this PR is ready for review, we need to decide on if we want to manage project tasks using a Makefile or using the pyproject.toml file with pixi. Currently, I've included both options, but maybe we should decide to go with one of them to avoid duplication? I've tried to list some pros/cons below:

**Makefile**
_Pros_
- Widely understood format
- Might be slightly easier to read compared to the pixi pyprojet.toml sections.

_Cons_
- Adds one more file to the repo
- Some might not be familiar with Makefiles

**pyproject.toml with pixi**
_Pros_
- Everything is kept in the pyproject.toml file - no additional files.
- Integrates well with pixi's support of multiple environments. (e.g. to run tests under python 3.12, one can run `pixi run -e py312-test test`. With the makefile, one can probably handle this by defining an optional argument to the make command)

_Cons_
- Slightly involved pixi sections, that maybe can be hard to read.

Do you have any opinion on the best path to go? @observingClouds, @trygveasp, @uandrae.

trygveasp:
I think I would prefer to have only a pyproject.toml file. Want to add @johtoblan to this as well since he has tested this pixi solution previously

observingClouds:
I have no strong feelings about this. The pixi tasks certainly allow for easier dynamic testing. What ever solution one chooses I would recommend to have it both in tactus and Deode-Workflow to let these repos not immediately diverge. Not sure what the vision of each of these is though.

mfroelund:
I cannot get the final tests to pass. I get the following errors:
```
ERROR tests/unit/test_commands_functions.py::test_namelist_integrate - FileNotFoundError: File /tmp/pytest-of-root/pytest-0/tactus/CY50t2_AROME_DEMO_60x80_2500m_20250209/20250209_0000/e165574babe71461/tactus/data/namelists/unit_testing/nl_master_integrate not found locally or relative to any sys.path entry
ERROR tests/unit/test_commands_functions.py::test_namelist_convert_yml - FileNotFoundError: File /tmp/pytest-of-root/pytest-0/tactus/CY50t2_AROME_DEMO_60x80_2500m_20250209/20250209_0000/e165574babe71461/tactus/data/namelists/unit_testing/nl_master_base.yml not found locally or relative to any sys.path entry
ERROR tests/unit/test_commands_functions.py::test_namelist_convert_ftn - FileNotFoundError: File /tmp/pytest-of-root/pytest-0/tactus/CY50t2_AROME_DEMO_60x80_2500m_20250209/20250209_0000/e165574babe71461/tactus/data/namelists/unit_testing/nl_master_base not found locally or relative to any sys.path entry
ERROR tests/unit/test_commands_functions.py::test_namelist_format_yml - FileNotFoundError: File /tmp/pytest-of-root/pytest-0/tactus/CY50t2_AROME_DEMO_60x80_2500m_20250209/20250209_0000/e165574babe71461/tactus/data/namelists/unit_testing/nl_master_base.yml not found locally or relative to any sys.path entry
ERROR tests/unit/test_commands_functions.py::test_namelist_format_ftn - FileNotFoundError: File /tmp/pytest-of-root/pytest-0/tactus/CY50t2_AROME_DEMO_60x80_2500m_20250209/20250209_0000/e165574babe71461/tactus/data/namelists/unit_testing/nl_master_base not found locally or relative to any sys.path entry
======= 559 passed, 1 skipped, 2 warnings, 5 errors in 215.15s (0:03:35) =======
```

All tests pass locally, and when running only the failed tests in the github action they pass, so it seems like there are some dependencies between the unit tests, which makes them fail. Cannot figure out what, and it's difficult to debug, since they only fail in github.. Any input is welcome!

observingClouds:
According to a "Colleague":
## Analysis of the failing tests

The root cause is that **all test modules share the same `getbasetemp()` directory**, and some tests (likely `test_cleaning.py` or `test_tasks.py`) delete or modify files in that shared directory before `test_commands_functions.py` runs — causing the `FileNotFoundError` for namelist files. The tests pass in isolation because no other module has touched the shared temp dir yet.

### Evidence

`tests/conftest.py` defines `tmp_directory` as:
```python
@pytest.fixture(scope="module")
def tmp_directory(tmp_path_factory):
    return tmp_path_factory.getbasetemp().as_posix()
```

`tmp_path_factory.getbasetemp()` returns the **same base temp path for the entire test session** — so all modules writing to or cleaning up from it share the same directory. The error path confirms this:
```
/tmp/pytest-of-root/pytest-0/tactus/CY50t2_.../tactus/data/namelists/unit_testing/nl_master_integrate
```
This is inside `getbasetemp()`, meaning a previous test wrote namelist files there and another test cleaned them up before `test_commands_functions.py` could use them.

### Recommended fix (Option 1): Use `mktemp()` for isolated per-module directories

Change `tests/conftest.py`:
```python
# BEFORE
@pytest.fixture(scope="module")
def tmp_directory(tmp_path_factory):
    """Return a temp directory valid for this module."""
    return tmp_path_factory.getbasetemp().as_posix()

# AFTER
@pytest.fixture(scope="module")
def tmp_directory(tmp_path_factory, request):
    """Return a temp directory valid for this module."""
    return tmp_path_factory.mktemp(request.module.__name__, numbered=False).as_posix()
```

And update the local override in `tests/unit/test_case_setup.py` to match (or remove it and let it inherit from `conftest.py`):
```python
# BEFORE
@pytest.fixture(scope="module", name="tmp_directory")
def fixture_tmp_directory(tmp_path_factory: pytest.TempPathFactory) -> Path:
    return tmp_path_factory.getbasetemp()

# AFTER — remove this fixture entirely, or:
@pytest.fixture(scope="module", name="tmp_directory")
def fixture_tmp_directory(tmp_path_factory: pytest.TempPathFactory, request) -> Path:
    return tmp_path_factory.mktemp(request.module.__name__, numbered=False)
```

This gives each test module its own isolated subdirectory, preventing cross-module contamination.

### Alternative quick fix (Option 2): Fix scope mismatch on `nlint_arg`

The `nlint_arg` fixture in `test_commands_functions.py` has default `function` scope but depends on `module`-scoped `tmp_directory`. Adding `scope="module"` may help stabilise it:
```python
@pytest.fixture(scope="module")
def nlint_arg(tmp_directory):
    ...
```

**Option 1 is the proper fix** — it eliminates the shared mutable state between modules entirely.

uandrae:
> I think I would prefer to have only a pyproject.toml file. Want to add @johtoblan to this as well since he has tested this pixi solution previously

Would agree with the pyproject.toml solution.

README.md
```diff
@@ -50,28 +45,51 @@ cd tactus
...
-  # Download and install poetry
-  curl -sSL https://install.python-poetry.org | python3 -
-  poetry install
-  # Add the poetry shell command as a plugin (for poetry >= v2.0.0)
-  poetry self add poetry-plugin-shell
+  pixi install
   ```
 
-Finally, install [`pygdal`](https://pypi.org/project/pygdal/), which is required for climate generation. [`pygdal`](https://pypi.org/project/pygdal/) depends on [`gdal`](https://gdal.org/), which is notoriously troublesome as dependency when targeting many systems. The versions of `pygdal` and the system's `gdal`should match.
+If you want to install the environment in another directory than the default (`<project-root>/.pixi/envs/`) add `detached-environments = "/path/to/env-location/"` to a pixi config file. Supported locations of the pixi config is outlined in https://pixi.prefix.dev/latest/reference/pixi_configuration/.
```
mfroelund, line 52: @jacobsnoeijer You should be able to use the below configuration to install the environment and cache at a custom location. Let me know how it works out on LUMI.

mfroelund:
Tests are now passing, so ready for review :)

trygveasp:
Failed to install on Fahrenheit:
```

sm_tryas@fahrenheit1:~/projects/ACCORD-NWP/tactus/mfroelund/feature/add-pixi$ pixi install
 WARN cache for Repodata at /home/sm_tryas/.cache/rattler/cache/repodata is on a network/parallel filesystem (NFS/SMB/FUSE/BeeGFS/Lustre/GPFS/CephFS), redirected to /tmp/pixi-cache-sm_tryas/repodata for this run. Set [cache.repodata] in config.toml or PIXI_CACHE_DIR to override, or [cache.netfs-redirect] = "never" to keep the original path.
 WARN cache for PypiMapping at /home/sm_tryas/.cache/rattler/cache/conda-pypi-mapping is on a network/parallel filesystem (NFS/SMB/FUSE/BeeGFS/Lustre/GPFS/CephFS), redirected to /tmp/pixi-cache-sm_tryas/conda-pypi-mapping for this run. Set [cache.conda-pypi-mapping] in config.toml or PIXI_CACHE_DIR to override, or [cache.netfs-redirect] = "never" to keep the original path.
▪ fetching repodata    [━━━━━━━━━━━━━━━━━━━━] 
▪ solving              [━━━━━━━━━━━━━━━━━━━━]  3/3                                                                                                                                                                                                                                                                                              
⠲ preparing packages   [━━━╾────────────────] 105/233 qt-main (+127) @ 17.2 MiB/s
⠴ installing           [━───────────────────] 88/233 xcb-util-renderutil (+16)                                                                                                                                                                                                                                                                  
Error:   × failed to solve the pypi requirements of environment 'py313' for platform 'linux-64'
  ╰─▶ build dispatch initialization failed: failed to link libboost-devel-1.91.0-hfcd1e18_0.conda

```

mfroelund:
> Failed to install on Fahrenheit:
> 
> ```
> 
> sm_tryas@fahrenheit1:~/projects/ACCORD-NWP/tactus/mfroelund/feature/add-pixi$ pixi install
>  WARN cache for Repodata at /home/sm_tryas/.cache/rattler/cache/repodata is on a network/parallel filesystem (NFS/SMB/FUSE/BeeGFS/Lustre/GPFS/CephFS), redirected to /tmp/pixi-cache-sm_tryas/repodata for this run. Set [cache.repodata] in config.toml or PIXI_CACHE_DIR to override, or [cache.netfs-redirect] = "never" to keep the original path.
>  WARN cache for PypiMapping at /home/sm_tryas/.cache/rattler/cache/conda-pypi-mapping is on a network/parallel filesystem (NFS/SMB/FUSE/BeeGFS/Lustre/GPFS/CephFS), redirected to /tmp/pixi-cache-sm_tryas/conda-pypi-mapping for this run. Set [cache.conda-pypi-mapping] in config.toml or PIXI_CACHE_DIR to override, or [cache.netfs-redirect] = "never" to keep the original path.
> ▪ fetching repodata    [━━━━━━━━━━━━━━━━━━━━] 
> ▪ solving              [━━━━━━━━━━━━━━━━━━━━]  3/3                                                                                                                                                                                                                                                                                              
> ⠲ preparing packages   [━━━╾────────────────] 105/233 qt-main (+127) @ 17.2 MiB/s
> ⠴ installing           [━───────────────────] 88/233 xcb-util-renderutil (+16)                                                                                                                                                                                                                                                                  
> Error:   × failed to solve the pypi requirements of environment 'py313' for platform 'linux-64'
>   ╰─▶ build dispatch initialization failed: failed to link libboost-devel-1.91.0-hfcd1e18_0.conda
> ```

Can you try to set the cache path to somewhere on home using https://github.com/ACCORD-NWP/tactus/pull/75#discussion_r3354552637?

observingClouds:
@trygveasp as we do not have access to that machine it would be helpful if you could provide additional information, e.g. by using `pixi install --verbose` and also try the other python environment with `pixi install --environment py312` and `pixi install --environment py314`.

trygveasp:
> @trygveasp as we do not have access to that machine it would be helpful if you could provide additional information, e.g. by using `pixi install --verbose` and also try the other python environment with `pixi install --environment py312` and `pixi install --environment py314`.

I have done it. The problem was actually reversed, as the file quota on HOME was full and I had to install on scratch.

It's now working on resolving 3 python versions which takes a bit of time.....

mfroelund:
> > @trygveasp as we do not have access to that machine it would be helpful if you could provide additional information, e.g. by using `pixi install --verbose` and also try the other python environment with `pixi install --environment py312` and `pixi install --environment py314`.
> 
> I have done it. The problem was actually reversed, as the file quota on HOME was full and I had to install on scratch.
> 
> It's now working on resolving 3 python versions which takes a bit of time.....

Actually there is no "py314" environment, but the default uses python 3.14, so if you only want to install that environment and not py312 and py313, you should run `pixi install -e default`. We could make a py314 environment if we'd like, but it would be identical to default.

docs/markdown_docs/local_installation.md
```diff
@@ -73,19 +73,19 @@ git clone git@github.com:ACCORD-NWP/tactus.git
...
+$ mamba create -p .conda ecflow gdal=3.5.0 python=3.10.4
 ...
 $ mamba activate .conda/
 ```
 
 Install tactus and all it's dependencies
 
 ```
-(tactus-py3.10) $ poetry install
+(tactus-py3.10) $ pixi install
```
observingClouds: As we now have a proper lock file, I suggest to install from that lock file by using `pixi install --frozen`. This way we ensure that dependencies are not randomly updated and everyone uses the exact same versions. In Deode-Workflow we have a dependabot action running that checks for updates periodically and opens PRs accordingly, such that version updates come in regularly and can be checked individually.

docs/markdown_docs/local_installation.md
```diff
@@ -73,19 +73,19 @@ git clone git@github.com:ACCORD-NWP/tactus.git
...
-Create a conda environment and install ecflow, gdal and poetry.
+Create a conda environment and install ecflow and gdal.
 ```
 $ module purge
 $ module load Mambaforge/23.3.1-1-hpc1
-$ mamba create -p .conda ecflow gdal=3.5.0 poetry python=3.10.4
+$ mamba create -p .conda ecflow gdal=3.5.0 python=3.10.4
 ...
 $ mamba activate .conda/
 ```
```
observingClouds: Is this necessary? Pixi handels now all dependencies including those available from conda-forge and creates a virtual environment.

pyproject.toml
```diff
@@ -82,13 +82,80 @@
...
+[tool.pixi.environments]
+  default = {features = ["py314"], solve-group = "py314"}
+  dev = {features = ["dev", "py314"], solve-group = "py314"}
+  doc = {features = ["doc", "py314"], solve-group = "py314"}
+  test = {features = ["test", "dev", "py314"], solve-group = "py314"}
+  py312 = {features = ["py312"], solve-group = "py312"}
+  py312-test = {features = ["test", "dev", "py312"], solve-group = "py312"}
+  py313 = {features = ["py313"], solve-group = "py313"}
+  py313-test = {features = ["test", "dev", "py313"], solve-group = "py313"}
+
```
observingClouds, line 132: It is probably good to have py314 defined here as well. Once support for 3.15 comes, it would definitely need to be created, but I think it is more natural to have it already.

README.md
observingClouds: Also no longer needed, right?! Should be covered by the pixi conda deps.

README.md
```diff
@@ -50,28 +45,51 @@ cd tactus
...
+
+Pixi caches the downloaded packages and share them between projects to speed up environment installation. To change the cache directory, add the following to the pixi config:
```
observingClouds:
```suggestion
Pixi caches the downloaded packages and shares them between projects to speed up environment installation. To change the cache directory, add the following to the pixi config:
```

README.md
```diff
@@ -50,28 +45,51 @@ cd tactus
...
+  root = "/path/to/cache"
+  netfs-redirect = "never"
+```
 
-  To install gdal and pygdal run the follow in commands in your shell:
+Finally, note that system `gdal` must be available on your system. The Pixi environment will install `pygdal` matching your system GDAL version automatically. If you need to specify the GDAL version, ensure it's installed via your system package manager:
 
   * On Atos (`hpc-login`)
     ```shell
     module load gdal/3.6.2
```
observingClouds: Also ships with pixi now

README.md
```diff
@@ -50,28 +45,51 @@ cd tactus
...
+    ```shell
+    conda create -n deode_env python=3.10 gdal ecflow
+    conda activate deode_env
+    pixi install
+    ```
+
+  * On LUMI
+    ```shell
+    module load crayPython
+    pixi install
```
observingClouds: As above.

trygveasp:
I have run this now on Fahrenheit until the MARS stuff which needs fixing. Pgd + input was fine. Modifcations needed:

```
[troika]
  troika = "/nobackup/forsk/sm_tryas/pixi/tactus-4397999228747316762/envs/default/bin/troika"

[submission.types.parallel]
  tasks = ["TactusBundleBuild", "Forecast", "TactusBuild", "C903"]

[submission.task_exceptions.TactusBundleBuild.ENV]
  OMP_NUM_THREADS = 144

[submission.task_exceptions.PgdUpdate]
  binary = "gl_DP"
```

And I needed to update the Fahrenheit toolchains, by adding:
```
set(ECBUILD_IFS_Fortran_FLAGS
    "-align" "array64byte"
    "-assume" "byterecl,realloc_lhs"
    "-traceback"
    "-fpic"
    "-qopenmp"
    "-qopenmp-threadprivate" "compat"
    "-fp-model" "precise"
    "-fp-speculation=safe"
    "-qopt-report=2"
    "-qopt-report-phase=all"
    "-fast-transcendentals"
    "-ftz"
    "-finline-functions"
    "-finline-limit=1500"
    "-Winline"
    "-diag-disable=5462"   # Global name too long, shortened symbol
    "-diag-disable=7713"   # disable unused statement functions warning
    "-diag-disable=11021"  # disable unresolved libraries in ipo warning
    "-diag-disable=10397"  # disable message reporting location of opt report files
    "-diag-disable=8291"   # disable remark: Recommended relationship between field width 'W' and the number of fractional digits 'D' in this edit descriptor is 'W>=D+7')
    "-diag-disable=10441") # disable ifort deprecation warning
```

Will set the boundaries complete and test with existing boundaries

trygveasp:
I was able to run this PR on Fahrenheit, but C903 still fails so needed to use pre-interpolated boundaries. Traceback:
```

forrtl: severe (174): SIGSEGV, segmentation fault occurred
Image              PC                Routine            Line        Source             
libc.so.6          000014B15D63FC30  Unknown               Unknown  Unknown
libectrans_dp.so   000014B1756E449B  gpnorm_trans_ctl_         179  gpnorm_trans_ctl_mod.F90
libectrans_lam_dp  000014B175A82A48  egpnorm_trans_dp_          97  egpnorm_trans.F90
MASTERODB          00000000071C052C  suemapf_                   92  suemapf.F90
MASTERODB          000000000058B890  sufpmapf_                  88  sufpmapf.F90
MASTERODB          0000000000585427  sufpgeometry_             271  sufpgeometry.F90
MASTERODB          0000000000545CCB  subfpos_                  312  subfpos.F90
MASTERODB          000000000044ED72  cprep3_                   197  cprep3.F90
MASTERODB          0000000000435E53  cnt0_                     198  cnt0.F90
MASTERODB          0000000000430104  MAIN__                    263  master.F90
MASTERODB          000000000042FA7D  Unknown               Unknown  Unknown
libc.so.6          000014B15D62A610  Unknown               Unknown  Unknown
libc.so.6          000014B15D62A6C0  __libc_start_main     Unknown  Unknown
MASTERODB          000000000042F995  Unknown               Unknown  Unknown
forrtl: severe (174): SIGSEGV, segmentation fault occurred
```

mfroelund:
Before I use to much time on it, I want to know if we still want to be able to install the project with poetry after merging this PR in? It should be possible, but doesn't work out-of-the-box, so it requires some more time to understand how to fix it.

README.md
```diff
@@ -50,28 +45,51 @@ cd tactus
...
+  * On Belenos (conda environment)
+    ```shell
+    conda create -n deode_env python=3.10 gdal ecflow
+    conda activate deode_env
+    pixi install
+    ```
+
+  * On LUMI
+    ```shell
+    module load crayPython
```
jacobsnoeijer: Module name should be something like `module load cray-python`

jacobsnoeijer:
**Status on LUMI**
When moving both `cache` and `venv` to `$SCRATCH` I managed to get tactus installed on LUMI. Getting a recent Python version seems working, getting the version number of a tool listed in the dependencies of tactus does also work but getting the version number from `tactus` results in a Segmentation Fault:

```
snoeijer@uan03:~/repos/ACCORD-NWP/tactus (pr74+75)$ pixi run python --version
Python 3.14.5
snoeijer@uan03:~/repos/ACCORD-NWP/tactus (pr74+75)$ pixi run tactus --version
tactus v1.1.0
```

_Note: in an earlier version of this comment a segmentation fault was mentioned; re-creating the environment results in a working Python 3.14 environment;_ 

Performance wise the setup on LUMI seems to be slow; it feels even slower as the previous Python3.10+poetry setup.

**Status on LEONARDO**
Setup a micromamba envirement with Python 3.12, etc. similar to the current way of installing on LEONARDO. With the python from this micromamba we can install tactus and its dependencies using pixi; running tactus with Python 3.14 works also on LEONARDO:
```
jsnoeije@login01:...ACCORD-NWP/tactus (pr74+75)$ pixi run python --version
Python 3.14.5
jsnoeije@login01:...ACCORD-NWP/tactus (pr74+75)$ pixi run tactus --version
tactus v1.1.0
```

pyproject.toml
```diff
@@ -3,18 +3,17 @@
   version = "1.1.0"
   description = "The Tactus scripting system."
   readme = "README.md"
-  requires-python = ">=3.10,<3.13"
   license = {text = "MIT"}
   authors = [{name = "ACCORD"}]
   keywords = ["forecasting", "tactus", "weather"]
+  requires-python = ">=3.12"
```
jacobsnoeijer: Can we weaken the requirement here to support python3.11 also? That's the latest system python on LUMI and the we can use that system python version to let pixi setup more recent environments for tactus

README.md
observingClouds: but this should not be needed at all with pixi

observingClouds:
@jacobsnoeijer how is the performance if you use the frozen environment, i.e. run `pixi install --frozen` so that pixi does not have to do the dependency resolution again. As pixi includes now also non-python dependencies and ships its own python this might take slightly longer, but not much longer.

observingClouds:
> Before I use to much time on it, I want to know if we still want to be able to install the project with poetry after merging this PR in? It should be possible, but doesn't work out-of-the-box, so it requires some more time to understand how to fix it.

I argue that the whole point of implementing pixi is to reduce the assumptions of the presence of non-pythonic dependencies ( e.g. `module load gdal`) and resolve them with pixi as well. This should allow us to remove quite a few `module load`s. The MPI library might be a different case for now, but certainly gdal and eccodes can come from conda-forge. I would therefore no longer support `poetry` and remove those `module load`s, particularly because the user should not even notice it much, just switching `poetry` for `pixi` in most cases.

docs/markdown_docs/local_installation.md
mfroelund: Maybe @uandrae or others can test if it works on freja without these commands ?

pyproject.toml
jacobsnoeijer: This comment can be ignored; we can keep the requirement python >=3.12

jacobsnoeijer:
> pixi install --frozen

`pixi install --frozen` does seem to improve performance although timings are quite slow:

For example `time pixi run -e py312 python --version` returns
```
Python 3.12.13

real    0m1,931s
user    0m0,388s
sys     0m1,423s
```

Next, `time pixi run -e py312 tactus --version` yields
```
tactus v1.1.0

real    3m46,986s
user    0m1,506s
sys     0m2,219s
```

**Note**: this ~4 minutes loading of the environment is sometimes also observable in task executions (i.e. consumed WALLTIME per task might increase with ~4 minutes on LUMI). This causes also the crash reported in https://github.com/ACCORD-NWP/tactus/pull/75#issuecomment-4649604342 when WALLTIME are set quite tight.

jacobsnoeijer:
For reference: to use the gdal loaded with pixi on LUMI one needs to remove
```
[submission.task_exceptions.Soil.MODULES]
  GDAL = ["load", "scl-gdal_23"]

[submission.task_exceptions.Topography.MODULES]
  GDAL = ["load", "scl-gdal_23"]

```
from `tactus/data/config_files/include/submission/lumi.toml`

Crash in Pgd due to WALLTIME: increased time from `0:10:00` to `0:20:00`

docs/markdown_docs/local_installation.md
uandrae: Works like a dream on freja! Only tested as far as the PgdInput to cover gdal usage and that went fine! No more mamba instructions required for freja.

README.md
uandrae: I suggest that lumi instructions are moved to local_installtion.md as it's not an official tactus platform.

jacobsnoeijer: Maybe even better to keep the lumi instructions in deode organization. But next to that loading module `cray-python` on LUMI should not be necessary anymore and can completely removed from the install instructions

docs/markdown_docs/local_installation.md
mfroelund: Will remove the section then

mfroelund: and make a common installation section and assume that this will also work for Belenos. @nicolasCtd can you test that by executing
```
curl -fsSL https://pixi.sh/install.sh | bash
(make sure to add pixi to path)
cd tactus
pixi install
pixi run tactus case <any-configuration-you-know-should-work> --start-suite
```
?

observingClouds:
Thanks for all the performance analysis @jacobsnoeijer 

The following comment strikes me
> pixi install --frozen does seem to improve performance although timings are quite slow

Do I interpret it correctly that once you have resolved the environment, e.g. with `pixi install --frozen` and you run some `tactus` tasks that the performance is degraded compared to when it was install with `poetry`? This would not make much sense, right? I mean these are both virtual environments that are (mostly) independent from the package manager once they are resolved. So I can only imagine that the python packages and python itself are installed on a partition that is not as performant compared to when installed with poetry.

docs/markdown_docs/local_installation.md
nicolasCtd: I get the following error:
```
Error:   × Cannot install environment 'default'
  ╰─▶ Wheel: eccodeslib-2.45.4.11-cp312-cp312-manylinux_2_28_x86_64.whl doesn't match this systems virtual capabilities for tags: cp312-cp312-manylinux_2_17_x86_64
```

But the same problem also occurs with Poetry... A workaround on Belenos is to remove eccodeslib from pyproject.toml and use a local installation of eccodes.

jacobsnoeijer:
> Thanks for all the performance analysis @jacobsnoeijer
> 
> The following comment strikes me
> 
> > pixi install --frozen does seem to improve performance although timings are quite slow
> 
> Do I interpret it correctly that once you have resolved the environment, e.g. with `pixi install --frozen` and you run some `tactus` tasks that the performance is degraded compared to when it was install with `poetry`? This would not make much sense, right? I mean these are both virtual environments that are (mostly) independent from the package manager once they are resolved. So I can only imagine that the python packages and python itself are installed on a partition that is not as performant compared to when installed with poetry.

My assumption is that reading from disk (or at least reading a lot of files from disk) is really slow om LUMI. Also with `poerty` loading the python environment before the first `tactus`-call was not always that fast but never in the order of 4 min. Before with poetry we used the system python / gdal / eccodes so I don't know if that can make a difference. At least the virtual environment with poerty was much smaller with <30K files while the `pixi` generated virtual environment doesn't fit into the 100K file limit.

uandrae:
On belenos I had to switch of the SSL certificat check by downlading the install.sh script and do some editing. Not very satisfactory. Would you be able to assist here @ludovicauger, @nicolasCtd, @tlestang ?

When running `pixi install` it stops on eccodes after a while

```
Error:   × Cannot install environment 'default'
  ╰─▶ Wheel: eccodeslib-2.45.4.11-cp314-cp314-manylinux_2_28_x86_64.whl doesn't match this systems virtual capabilities for tags:
      cp314-cp314-manylinux_2_17_x86_64
      cp314-cp314-manylinux2014_x86_64
...
```
Does anyone have any knowledge about python versions and eccodes/eccodeslib on belenos?

observingClouds:
> doesn't match this systems virtual capabilities for tags: cp314-cp314-manylinux_2_17_x86_64

So this means Belenos only advertises compatibility with manylinux_2_17 wheels, while this package requires manylinux_2_28. Since manylinux_2_17 corresponds to glibc 2.17, it suggests the system userspace is based on a fairly old glibc baseline from 2012 (!). 2_28 was released in 2018 so it is not an outrages requirement :D

uandrae:
> > doesn't match this systems virtual capabilities for tags: cp314-cp314-manylinux_2_17_x86_64
> 
> So this means Belenos only advertises compatibility with manylinux_2_17 wheels, while this package requires manylinux_2_28. Since manylinux_2_17 corresponds to glibc 2.17, it suggests the system userspace is based on a fairly old glibc baseline from 2012 (!). 2_28 was released in 2018 so it is not an outrages requirement :D

Indeed
```
[andraeu@belenoslogin1 ~]$ ldd --version
ldd (GNU libc) 2.17
```
Regardless of what we think targeted platforms should look like we need to handle these cases. Especially for a platform like belenos which is considered as a platform where tactus should work. What's the level of configurability with pixi given that we have been able to install with micromamba/conda?

nicolasCtd:
> On belenos I had to switch of the SSL certificat check by downlading the install.sh script and do some editing. Not very satisfactory. Would you be able to assist here @ludovicauger, @nicolasCtd, @tlestang ?
> 
You have to use the certificate: `export CURL_CA_BUNDLE="/opt/softs/certificats/proxy1_1.pem"`

docs/markdown_docs/local_installation.md
mfroelund: Belenos discussion continues in https://github.com/ACCORD-NWP/tactus/pull/75#issuecomment-4652487142

observingClouds:
@uandrae or others with access to Belenos, could you try removing the `eccodes` and `eccodeslib` dependencies from `[dependencies]`  in `pyproject.toml` and add `python-eccodes = "*"` to `[tool.pixi.dependencies]`? This way we omit the python wheels and might not have the glib restriction.

uandrae:
> @uandrae or others with access to Belenos, could you try removing the `eccodes` and `eccodeslib` dependencies from `[dependencies]` in `pyproject.toml` and add `python-eccodes = "*"` to `[tool.pixi.dependencies]`? This way we omit the python wheels and might not have the glib restriction.

Thanks for the suggestion. Tried it but I'm afraid eccodes is not the only library depending on newer versions. This is the next problem:
```
Error:   × Cannot install environment 'default'
  ╰─▶ Wheel: pyproj-3.7.2-cp314-cp314-manylinux_2_28_x86_64.whl doesn't match this systems virtual capabilities for tags: cp314-cp314-manylinux_2_17_x86_64
```

observingClouds:
Well, as container support will probably still take some time, we could check if all those wheels that Belenos has issues with are available on conda-forge. For `pyproj` this is for example also the case, so you could move `pyproj` also from `[dependencies]` to `[tool.pixi.dependencies]`. If this is of interest I would ask you to do this game until the environment resolves and then report the packages here.

uandrae:
> Well, as container support will probably still take some time, we could check if all those wheels that Belenos has issues with are available on conda-forge. For `pyproj` this is for example also the case, so you could move `pyproj` also from `[dependencies]` to `[tool.pixi.dependencies]`. If this is of interest I would ask you to do this game until the environment resolves and then report the packages here.

Thanks for the suggestion. With the changes below we arrive to

```
(tactus) (tactus) [andraeu@belenoslogin0 tactus]$ tactus --version
2026-06-12 08:53:55 | INFO     | Start tactus v1.1.0 --> "tactus --version"
tactus v1.1.0
(tactus) (tactus) [andraeu@belenoslogin0 tactus]$ python3 --version
Python 3.14.5
```

Of course the proof of functionality is in the full execution, but at least this gives a hint about possibilities and limitations.

```
(tactus) (tactus) [andraeu@belenoslogin0 tactus]$ git diff pyproject.toml
diff --git a/pyproject.toml b/pyproject.toml
index 65fb5f73..ad69ae01 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,8 +21,8 @@
     "dateutils>=0.6.12",
     "dicttoxml>=1.0.0",
     "ecbundle==2.4.0",
-    "eccodes==2.45",
-    "eccodeslib==2.45.4.11",
+    #"eccodes==2.45",
+    #"eccodeslib==2.45.4.11",
     "f90nml>=1.4.3",
     "fastjsonschema==2.18.1",
     "frozendict>=2.4.7",
@@ -33,16 +33,16 @@
     "humanize>=4.8.0",
     "isodate>=0.6.1",
     "json-schema-for-humans>=1.5.1",
-    "json2tab @ git+https://github.com/mfroelund/json2tab.git@feature/adjust-python-version",
+    #"json2tab @ git+https://github.com/mfroelund/json2tab.git@feature/adjust-python-version",
     "jsonref>=1.1.0",
     "loguru>=0.7.1",
-    "numexpr>=2.8.4",
+    #"numexpr>=2.8.4",
     "numpy>=1.22.4",
     "omegaconf>=2.3",
-    "pandas>=1.4.0,<3.0.0",
+    #"pandas>=1.4.0,<3.0.0",
     "pydantic>=2.9.2",
     "pyfdb==0.1.3",
-    "pyproj>=3.3.0,!=3.5.0",
+    #"pyproj>=3.3.0,!=3.5.0",
     "pysurfex>=0.1.2",
     "pyyaml>=6.0",
     "thenamelisttool @ git+https://github.com/UMR-CNRM/thenamelisttool.git",
@@ -102,7 +102,11 @@
 
 [tool.pixi.dependencies]
   # conda-forge provides C libraries that pip wheels don't bundle
+  numexpr = "*"
+  pandas = "*"
+  pyproj = "*"
   pip = "*"
+  python-eccodes = "*"
   eccodes = "==2.45"
   ecflow = ">=5.15.0"
   gdal = "*"
```

mfroelund:
> > Well, as container support will probably still take some time, we could check if all those wheels that Belenos has issues with are available on conda-forge. For `pyproj` this is for example also the case, so you could move `pyproj` also from `[dependencies]` to `[tool.pixi.dependencies]`. If this is of interest I would ask you to do this game until the environment resolves and then report the packages here.
> 
> Thanks for the suggestion. With the changes below we arrive to
> 
> ```
> (tactus) (tactus) [andraeu@belenoslogin0 tactus]$ tactus --version
> 2026-06-12 08:53:55 | INFO     | Start tactus v1.1.0 --> "tactus --version"
> tactus v1.1.0
> (tactus) (tactus) [andraeu@belenoslogin0 tactus]$ python3 --version
> Python 3.14.5
> ```
> 
> Of course the proof of functionality is in the full execution, but at least this gives a hint about possibilities and limitations.
> 
> ```
> (tactus) (tactus) [andraeu@belenoslogin0 tactus]$ git diff pyproject.toml
> diff --git a/pyproject.toml b/pyproject.toml
> index 65fb5f73..ad69ae01 100644
> --- a/pyproject.toml
> +++ b/pyproject.toml
> @@ -21,8 +21,8 @@
>      "dateutils>=0.6.12",
>      "dicttoxml>=1.0.0",
>      "ecbundle==2.4.0",
> -    "eccodes==2.45",
> -    "eccodeslib==2.45.4.11",
> +    #"eccodes==2.45",
> +    #"eccodeslib==2.45.4.11",
>      "f90nml>=1.4.3",
>      "fastjsonschema==2.18.1",
>      "frozendict>=2.4.7",
> @@ -33,16 +33,16 @@
>      "humanize>=4.8.0",
>      "isodate>=0.6.1",
>      "json-schema-for-humans>=1.5.1",
> -    "json2tab @ git+https://github.com/mfroelund/json2tab.git@feature/adjust-python-version",
> +    #"json2tab @ git+https://github.com/mfroelund/json2tab.git@feature/adjust-python-version",
>      "jsonref>=1.1.0",
>      "loguru>=0.7.1",
> -    "numexpr>=2.8.4",
> +    #"numexpr>=2.8.4",
>      "numpy>=1.22.4",
>      "omegaconf>=2.3",
> -    "pandas>=1.4.0,<3.0.0",
> +    #"pandas>=1.4.0,<3.0.0",
>      "pydantic>=2.9.2",
>      "pyfdb==0.1.3",
> -    "pyproj>=3.3.0,!=3.5.0",
> +    #"pyproj>=3.3.0,!=3.5.0",
>      "pysurfex>=0.1.2",
>      "pyyaml>=6.0",
>      "thenamelisttool @ git+https://github.com/UMR-CNRM/thenamelisttool.git",
> @@ -102,7 +102,11 @@
>  
>  [tool.pixi.dependencies]
>    # conda-forge provides C libraries that pip wheels don't bundle
> +  numexpr = "*"
> +  pandas = "*"
> +  pyproj = "*"
>    pip = "*"
> +  python-eccodes = "*"
>    eccodes = "==2.45"
>    ecflow = ">=5.15.0"
>    gdal = "*"
> ```

These changes will most likely also work on other machines than Belenos, so should I implement them on the branch, and then we can do a full test on Belenos to see if it works?

nicolasCtd:
> > Well, as container support will probably still take some time, we could check if all those wheels that Belenos has issues with are available on conda-forge. For `pyproj` this is for example also the case, so you could move `pyproj` also from `[dependencies]` to `[tool.pixi.dependencies]`. If this is of interest I would ask you to do this game until the environment resolves and then report the packages here.
> 
> Thanks for the suggestion. With the changes below we arrive to
> 
> ```
> (tactus) (tactus) [andraeu@belenoslogin0 tactus]$ tactus --version
> 2026-06-12 08:53:55 | INFO     | Start tactus v1.1.0 --> "tactus --version"
> tactus v1.1.0
> (tactus) (tactus) [andraeu@belenoslogin0 tactus]$ python3 --version
> Python 3.14.5
> ```
> 
> Of course the proof of functionality is in the full execution, but at least this gives a hint about possibilities and limitations.
> 
> ```
> (tactus) (tactus) [andraeu@belenoslogin0 tactus]$ git diff pyproject.toml
> diff --git a/pyproject.toml b/pyproject.toml
> index 65fb5f73..ad69ae01 100644
> --- a/pyproject.toml
> +++ b/pyproject.toml
> @@ -21,8 +21,8 @@
>      "dateutils>=0.6.12",
>      "dicttoxml>=1.0.0",
>      "ecbundle==2.4.0",
> -    "eccodes==2.45",
> -    "eccodeslib==2.45.4.11",
> +    #"eccodes==2.45",
> +    #"eccodeslib==2.45.4.11",
>      "f90nml>=1.4.3",
>      "fastjsonschema==2.18.1",
>      "frozendict>=2.4.7",
> @@ -33,16 +33,16 @@
>      "humanize>=4.8.0",
>      "isodate>=0.6.1",
>      "json-schema-for-humans>=1.5.1",
> -    "json2tab @ git+https://github.com/mfroelund/json2tab.git@feature/adjust-python-version",
> +    #"json2tab @ git+https://github.com/mfroelund/json2tab.git@feature/adjust-python-version",
>      "jsonref>=1.1.0",
>      "loguru>=0.7.1",
> -    "numexpr>=2.8.4",
> +    #"numexpr>=2.8.4",
>      "numpy>=1.22.4",
>      "omegaconf>=2.3",
> -    "pandas>=1.4.0,<3.0.0",
> +    #"pandas>=1.4.0,<3.0.0",
>      "pydantic>=2.9.2",
>      "pyfdb==0.1.3",
> -    "pyproj>=3.3.0,!=3.5.0",
> +    #"pyproj>=3.3.0,!=3.5.0",
>      "pysurfex>=0.1.2",
>      "pyyaml>=6.0",
>      "thenamelisttool @ git+https://github.com/UMR-CNRM/thenamelisttool.git",
> @@ -102,7 +102,11 @@
>  
>  [tool.pixi.dependencies]
>    # conda-forge provides C libraries that pip wheels don't bundle
> +  numexpr = "*"
> +  pandas = "*"
> +  pyproj = "*"
>    pip = "*"
> +  python-eccodes = "*"
>    eccodes = "==2.45"
>    ecflow = ">=5.15.0"
>    gdal = "*"
> ```

I made these changes on Belenos. I also managed to create the environment with Pixi, but `json2tab` is missing, so I get the following error when I launch an experiment:
```
2026-08-27 10:28:30 | INFO     | Loading module tactus.tasks.generatewfptabfile
2026-08-27 10:28:30 | ERROR    | Could not load tactus.tasks.generatewfptabfile: ModuleNotFoundError("No module named 'json2tab'")
Traceback (most recent call last):
  File "/home/gmap/mrpa/coutandn/pixi/test/tactus/tactus/tasks/discover_task.py", line 42, in discover_modules
    mod = importlib.import_module(fullname)
  File "/home/gmap/mrpa/coutandn/pixi/test/tactus/.pixi/envs/default/lib/python3.14/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1406, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1371, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1342, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 938, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 759, in exec_module
  File "<frozen importlib._bootstrap>", line 491, in _call_with_frames_removed
  File "/home/gmap/mrpa/coutandn/pixi/test/tactus/tactus/tasks/generatewfptabfile.py", line 5, in <module>
    from json2tab import json2tab
```
When I try to add json2tab, the dependencies cannot be resolved.
```
Error:   × failed to solve the pypi requirements of environment 'py313' for platform 'linux-64'
  ├─▶ failed to resolve pypi dependencies
  ╰─▶ Because json2tab==0.7.0 depends on pandas>=1.4.0,<3.0.0 and pandas==3.0.5, we can conclude that json2tab==0.7.0 cannot be used.
      And because only json2tab==0.7.0 is available and you require json2tab, we can conclude that your requirements are unsatisfiable.
```
What still works for me on Belenos is to remove `eccodes` and `eccodeslib` dependecies, and use a seperate installation of eccodes by setting environment variables.

uandrae:
What's the situation for lumi & leonardo when it comes to moving forward with python versions @jacobsnoeijer?

jacobsnoeijer:
Nothing changed; also haven't looked into it after summer

PR #88, opened 2026-05-31
Use mirrors in test-runner

uandrae:
## Describe your changes
Introduces the possibility to reuse data from different suites to reduce the overall cost. The launch is now separated in two steps

- `-m` to create the config files
- `-r` to launch the suites

The generation of config files is now done in parallel wich reduces the configuration time considerably. The threading can we switched off by setting `general.max_workers=1`.

Ready for testing but still needs some documentation updates.

< Please also include relevant motivation and context. >

< List any dependencies that are required for this change. >

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Testing

- [ ] I have tested this on ATOS using the `atos_bologna.toml` configuration in the latest version of [Tactus-test-runner](https://github.com/destination-earth-digital-twins/Tactus-test-runner)
- [ ] I have tested this on LUMI using the `lumi.toml` configuration in the latest version of [Tactus-test-runner](https://github.com/destination-earth-digital-twins/Tactus-test-runner)

For further information see the [development guide](https://github.com/ACCORD-NWP/tactus/blob/develop/docs/markdown_docs/development_guide.md)

### Code quality

- [ ] My change follows the [best practices for this project](https://github.com/ACCORD-NWP/tactus/blob/develop/docs/markdown_docs/development_guide.md#best-practices).
- [ ] My local environment is correctly initialised as described in the [README](https://github.com/ACCORD-NWP/tactus/blob/develop/README.md) file.
- [ ] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the documentation and docstrings to reflect the changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have ensured that the code is still installable with `poetry` after the changes and runs
- [ ] I have requested one or more reviewer(s) and an assignee (assignee is responsible for merging). At least one reviewer has accepted to review.

## Checklist for reviewers
Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [ ] the code readable
- [ ] the code well tested (checked coverage report)
- [ ] the code documented
- [ ] the code easy to maintain

## Author checklist after completed review

[...]

PR #95, opened 2026-06-03
Deactivate the cache for compilation by default

dhaumont:
## Describe your changes

Deactivate the cache mechanism for compilation, such that the test suite works out of the box for CY50.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

- [x] I have added a line to the CHANGELOG describing this change (in section
  missing)

## Checklist for assignees
- [ ] the tests passing

## Tag possible reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/destination-earth-digital-twins/Deode-Prototype/tree/HEAD)
 
 ### Added
+- Deactivate the cache for compilation by default[#95](https://github.com/ACCORD-NWP/tactus/pull/95) (@dhaumont)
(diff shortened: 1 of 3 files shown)


pardallio:
Shouldn t it be deactivated just for the test suites then?

pardallio:
I think @uandrae was going to add a compilation only suite to the test bed so that other suites would use its binaries for testing

pardallio:
Of course, if there are other reasons to deactivate ot other than testing, we should consider it

uandrae:
> I think @uandrae was going to add a compilation only suite to the test bed so that other suites would use its binaries for testing

Yup, this is done in #88. I think this PR handles the case where users naively starts a number of cases, something I thing should be allowed.

pardallio (approved):
Lgtm!

dhaumont:
> Yup, this is done in #88. I think this PR handles the case where users naively starts a number of cases, something I thing should be allowed.

That's the thing. Naive users can't use it as it is now :-)

PR #101, opened 2026-06-04
Introduce end relative to start

uandrae:
## Describe your changes
Changes the convention for  `general.times.end` to make it relative to `general.times.start` when express in duration notation. The background is the need in testing over more than one cycle and where the start date is not known. 

The need for this functionality will be introduced in #88. 

< Please also include relevant motivation and context. >

< List any dependencies that are required for this change. >

## Type of change

### Testing

### Code quality

## Checklist for reviewers
- [ ] the code readable
- [ ] the code documented

  missing)

## Checklist for assignees
- [ ] the tests passing

## Tag possible reviewers

### DIFF
--- a/tactus/data/config_files/config_file_schemas/main_config_schema.json
+++ b/tactus/data/config_files/config_file_schemas/main_config_schema.json
@@ -901,7 +901,7 @@
                 },
                 {
                   "format": "duration",
-		  "pattern": "^P"
+                  "pattern": "^P"
                 }
               ]
             },
--- a/tactus/data/config_files/config_file_schemas/main_config_schema.json
+++ b/tactus/data/config_files/config_file_schemas/main_config_schema.json
@@ -881,6 +881,9 @@
                 },
                 {
                   "format": "duration"
+                },
+                {
+                  "pattern": "^relative_to_start(.*)"
                 }
               ]
             },
--- a/tactus/data/config_files/config_file_schemas/main_config_schema.json
+++ b/tactus/data/config_files/config_file_schemas/main_config_schema.json
@@ -904,8 +904,7 @@
                   "format": "date-time"
                 },
                 {
-                  "format": "duration",
-                  "pattern": "^P"
+                  "format": "duration"
                 }
               ]
             },
--- a/tactus/data/config_files/config_file_schemas/main_config_schema.json
+++ b/tactus/data/config_files/config_file_schemas/main_config_schema.json
@@ -893,17 +893,15 @@
               ]
             },
             "end": {
-              "description": "End time if the run covers several cycles. Same format as start",
+              "description": "End time if the run covers several cycles. Date time format or duration relative start.",
               "type": "string",
               "oneOf": [
                 {
                   "format": "date-time"
                 },
                 {
-                  "format": "duration"
-                },
-                {
-                  "pattern": "^relative_to_start(.*)"
+                  "format": "duration",
+		  "pattern": "^P"
                 }
               ]
             },
--- a/tactus/datetime_utils.py
+++ b/tactus/datetime_utils.py
@@ -249,7 +249,7 @@ def get_month_list(start, end) -> list:
     ]
 
 
-def evaluate_date(date: str) -> str:
+def evaluate_date(date: str, reference_date=None) -> str:
     """Parses an ISO 8601 datetime and/or duration and returns the computed datetime.
 
     - If the input is a datetime (e.g., "2025-03-19T00:00:00Z"), it returns it as-is.
@@ -260,6 +260,7 @@ def evaluate_date(date: str) -> str:
 
     Args:
         date (str): An ISO 8601 datetime, duration, or both.
+        reference_date (str): An ISO 8601 datetime, duration, or both.
 
     Returns:
         str: The computed datetime in ISO format.
@@ -276,11 +277,14 @@ def evaluate_date(date: str) -> str:
         )
 
     if date.startswith(("P", "-P")):
-        today_midnight = datetime.now(timezone.utc).replace(
-            hour=0, minute=0, second=0, microsecond=0
-        )
+        if reference_date is None:
+            _reference_date = datetime.now(timezone.utc).replace(
+                hour=0, minute=0, second=0, microsecond=0
+            )
+        else:
+            _reference_date = as_datetime(evaluate_date(reference_date))
         return (
-            (today_midnight + isodate.parse_duration(date))
+            (_reference_date + isodate.parse_duration(date))
             .isoformat(timespec="seconds")
             .replace("+00:00", "Z")
         )
--- a/tactus/datetime_utils.py
+++ b/tactus/datetime_utils.py
@@ -278,9 +278,10 @@ def evaluate_date(date: str, reference_date=None) -> str:
 
     if date.startswith(("P", "-P")):
         if reference_date is None:
-            _reference_date = datetime.now(timezone.utc).replace(
+            today_midnight = datetime.now(timezone.utc).replace(
                 hour=0, minute=0, second=0, microsecond=0
             )
+            _reference_date = today_midnight
         else:
             _reference_date = as_datetime(evaluate_date(reference_date))
         return (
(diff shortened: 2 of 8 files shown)


mfroelund:
Do we already have the possibility to set the end datetime as a duration with a different purpose of the one implemented in this PR? If not, couldn't we just skip the "relative_to_start" prefix, and have the convention, that a end duration is always relative to the start? Or will there be other cases where the relative to is different?

uandrae:
Currently `general.times.end` is defined as `End time if the run covers several cycles. Same format as start`. I.e. I don't think we should change it. We could implement `general.times.relative_end`  which could override end if set. A bit cleaner compared to what I've done in this PR.

uandrae:
Actually I think your suggestion is better so I've implemented this. I don't think it's a big deal at this stage to change the meaning of end. The json schema will capture potential old habits.

uandrae:
Bad idea, more work to do...

uandrae:
Back in buisness and works as expected.

observingClouds:
I think it is a really good point to define the end time relative to the start time (at least optionally) as the same key (`general.times.start`). I want to emphasize that these config files really need to be versioned though and that changes need to be documented. It is otherwise really hard for any tooling that builds on top to maintain compatibility as it basically needs to guess.

uandrae:
How do you suggest to version the config files?

uandrae:
> How do you suggest to version the config files?

Or what do we need beoynd the standard tagging @observingClouds?

uandrae:
Will you have time to do a second review @mfroelund?

tests/unit/test_derived_variables.py
```diff
@@ -5,17 +5,30 @@
 
 import pytest
 
+from tactus.config_parser import ConfigFileValidationError
 from tactus.derived_variables import derived_variables, set_times
 
 
-def test_set_times(default_config):
```
dhaumont, line 11: Shouldn't we keep the `test_set_times` test ?

tactus/datetime_utils.py
```diff
@@ -276,11 +277,14 @@ def evaluate_date(date: str) -> str:
         )
 
     if date.startswith(("P", "-P")):
-        today_midnight = datetime.now(timezone.utc).replace(
-            hour=0, minute=0, second=0, microsecond=0
-        )
+        if reference_date is None:
```
dhaumont, line 280: for code lisibility, I would keep the `today_midnight` local variable: 
```
 if reference_date is None:
    today_midnight = datetime.now(timezone.utc).replace(
                 hour=0, minute=0, second=0, microsecond=0
             )
    _reference_data = today_midnight
```

tactus/derived_variables.py
```diff
@@ -38,12 +38,13 @@ def set_times(config):
...
         times.update({"start": times["basetime"]})
         logger.debug("Set start to {}", times["start"])
+
     if "end" not in times:
         times.update({"end": times["basetime"]})
         logger.debug("Set end to {}", times["end"])
 
     times.update({"start": evaluate_date(times["start"])})
-    times.update({"end": evaluate_date(times["end"])})
+    times.update({"end": evaluate_date(times["end"], reference_date=times["start"])})
```
dhaumont, line 47: Do I understand correctly that you change the behaviour here, and always force end to be  relative to start ? Is it what we want ?

tactus/derived_variables.py
```diff
@@ -195,8 +196,12 @@ def derived_variables(config, processor_layout=None):
         )
         raise NotImplementedError(msg)
 
-    xlat0 = config.get("domain.xlat0", config.get("domain.xlatcen"))
-    xlon0 = config.get("domain.xlon0", config.get("domain.xloncen"))
+    xlat0 = config.get("domain.xlat0", "")
```
dhaumont: I believe this is not related to this PR, but the changes looks okay.

tactus/derived_variables.py
```diff
@@ -195,8 +196,12 @@ def derived_variables(config, processor_layout=None):
...
+    xlon0 = config.get("domain.xlon0", "")
+    if not xlat0:
+        xlat0 = config.get("domain.xlatcen")
+    if not xlon0:
+        xlon0 = config.get("domain.xloncen")
```
dhaumont, line 201: Do we want to add a warning or a default value if xlat0 or xlon0 is still empty after L204 ?

tactus/experiment.py
```diff
@@ -57,7 +57,12 @@ def __init__(
             config = config.copy(
                 update={
                     "general": {
-                        "times": {"end": evaluate_date(config["general.times.end"])}
+                        "times": {
+                            "end": evaluate_date(
+                                config["general.times.end"],
+                                reference_date=config["general.times.start"],
```
dhaumont, line 63: Same here, the end is always relative. Is this fine ?

tactus/toolbox.py
```diff
@@ -609,7 +609,8 @@ def substitute(
 
             end = self.config.get("general.times.end", None)
             if end is not None:
-                pattern = self.substitute_datetime(pattern, as_datetime(end), "_END")
+                _end = evaluate_date(end, reference_date=start)
+                pattern = self.substitute_datetime(pattern, as_datetime(_end), "_END")
```
dhaumont, line 613: I'm surprised that you evaluate the pattern after evaluate_date, I was expecting to resolve the pattern before. But I don't have all the logic in mind

dhaumont (commented):
In general, the code changes look fine. I just added a few questions and minor suggestions.

tactus/toolbox.py
uandrae: This is to resolved macros with e.g. @YYYYMMDD_END@ and for this we first have to translate e.g. `end ="P1D"` to a datetime string. When this is done we can do the substitution.

tactus/derived_variables.py
uandrae: Merge error, will fix.

tactus/derived_variables.py
uandrae: We force end to be relative start when we define it as a timedelta object. You can still define end explicitly. Note that if end is not set it will always be equal to start. Possible current usage of e.g. `end = "-P1D"` will fail as we do not allow the timedelta to be negative since it implies end < start.

tactus/experiment.py
uandrae: See above

tests/unit/test_derived_variables.py
mfroelund, line 1: I suggest to add a test for the positive case with a valid start date and end duration

mfroelund (commented):
Looks good, small comment for the testing.

tests/unit/test_derived_variables.py
dhaumont: +1

tactus/data/config_files/config_file_schemas/main_config_schema.json
```diff
@@ -897,14 +897,15 @@
...
+              "description": "End time if the run covers several cycles. Date time format or duration relative start.",
               "type": "string",
               "oneOf": [
                 {
                   "format": "date-time"
                 },
                 {
-                  "format": "duration"
+                  "format": "duration",
+                  "pattern": "^P"
```
mfroelund: We can keep this if you'd like, but it's not strictly necessary since we already have https://github.com/ACCORD-NWP/tactus/pull/101/changes#diff-fafe7d84f85ccbedc0b388b844255eaf96782b5001ebb7180f52d70b0c9893a5R49 .
One argument for removing "^P" is, that the error message from the json schema validation is not as informative:
"tactus.config_parser.ConfigFileValidationError: "general -> times -> end" must be valid exactly by one definition (0 matches found). Received type "str" with value "-PT5H"."

as the above mentioned exception:
"ValueError: general.times.start (2026-06-10T14:00:00Z) cannot be larger than general.times.end (2026-06-10T09:00:00Z)"

uandrae: Good point, will remove it.

uandrae: Done!

mfroelund (approved):

dhaumont (approved):

PR #111, opened 2026-06-12
Implement usage of the reference checker

dhaumont:
## Describe your changes

This PR implements the usage of the reference checker in tactus test, see https://github.com/destination-earth-digital-twins/Deode-Workflow/pull/1524.
It is the continuation of the tactus-test-runner PR https://github.com/destination-earth-digital-twins/Tactus-test-runner/pull/32

The following changes are introduced:
- add a new optional `--generate-references` or `-g` argument to the `tactus test` command to generate the reference files

``` 
tactus test -c  tactus/data/tests/atos_bologna_CY50t2.toml -g
```
- fix the reference_date of the tests to `2025-02-09T00:00:00Z` (instead of the day before) to have a reference stored to compare with.

- activate the reference checks against a stored reference for all the test suites. The following files are tested:
    - ICMSH.*0003h00m00s grib file
    - ICMSH.*0003h00m00s FA file
    - NODE_01.01 file
    - ICMSH.*0003h00m00s.sfx FA file
This is using a new rule that was introduced to test surfex output.
The threshold for norms was set to zero (no difference allowed) to detect bit non-reproductibility-
- <del>add the @PRECISION@ to the name of the tests
`feature_activate_reference_checker_intel_R64_CY50t2_HARMONIE_AROME_DEMO_60x80_2500m_20250209_mode_start` instead of `feature_activate_reference_checker_intel_CY50t2_HARMONIE_AROME_DEMO_60x80_2500m_20250209_mode`_start`</del>
- in reference_checker:
   -  verify that all parameters are correctly defined for the rules
   - add an error_message mechanism to report errors during analysis
   - automatically delete the summary file if a computation is restarted but the summary was already present and complete (ie contained a summary section in the end)
   - always create the ReferenceCheck task, even for the `*_climate` suites
- introduce `modifs_reference_checker.toml` and `modifs_reference_generate.toml` in `data/tests`, and merge these new dictionaries in test_runner.
- add the possibility to display a summary of the test by typing the test command without argument

```
$ tactus test -c tactus/data/tests/atos_bologna_CY50t2.toml  
[..]
2026-06-12 15:01:31 | INFO     |  tag: feature_activate_reference_checker_
2026-06-12 15:01:31 | INFO     |  test_dir: feature_activate_reference_checker_configs
2026-06-12 15:01:31 | INFO     | Comparison against /scratch/cvap/tactus_references/v1.1.0
[...]

PR #128, opened 2026-06-19
ReferenceChecker: Improve the way the summary are created

dhaumont:
## Describe your changes

When you run a suite with reference checker, a summary file is created, containing a list of all the tests performed and their results.
In the previous implementation, the summary is created once during the 'Prep' task, each test append data to it, and append a final analysis to the summary a the last ReferenceCheck task.

This works well, until you want to restart some part of the suite without executing 'Prep' : the summary is not deleted or recreated, and the new tests are appended to the previous summary.

This PR solves this issue, by deleting the summary if the analysis is already present in it.
It also adds some sanity check to make sure that the configuration is correct and that all the parameters of the rules are correctly defined.

Note: This has been extracted from #111.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

### Testing

### Code quality

## Checklist for reviewers
- [ ] the code readable
- [ ] the code documented

  missing)

## Checklist for assignees
- [ ] the tests passing

## Tag possible reviewers

### DIFF
--- a/tactus/reference_checker.py
+++ b/tactus/reference_checker.py
@@ -351,6 +351,27 @@ class CheckDefinition:
         if taskname in config["task"]:
             for rulename in rules_active:
                 if rulename in config["task"][taskname]:
+                    parameters = ["method", "inpath", "pattern", "result_folder"]
+                    if generate:
+                        parameters.append("generate_folder")
+                    if check:
+                        parameters.append("reference_folder")
+
+                    have_all_parameters = True
+                    for parameter in parameters:
+                        if parameter not in config["task"][taskname][rulename]:
+                            logger.warning(
+                                f"Reference Checker - {parameter} not defined for"
+                                + f"task {taskname} and rule {rulename}."
+                            )
+                            have_all_parameters = False
+                    if not have_all_parameters:
+                        logger.warning(
+                            f"Skipping reference check definition for {taskname}"
+                            + f" and rule {rulename}"
+                        )
+                        continue
+
                     method = config["task"][taskname][rulename]["method"]
                     inpath = config["task"][taskname][rulename]["inpath"]
                     pattern = config["task"][taskname][rulename]["pattern"]
@@ -499,6 +520,7 @@ class CheckSummaryAnalysis:
         self.missing_count = 0
         self.generated_count = 0
         self.check = check
+        self.error_message = ""
 
     def increment_error_count(self):
         """Increment error_count."""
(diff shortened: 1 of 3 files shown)


dhaumont:
@kastelecn @pardallio Hello Nika and Miguel. Do you want to review this one ? If not, can you tag someone who could ? Thanks !

tactus/reference_checker.py
```diff
@@ -895,6 +920,14 @@ def compute_and_append_analysis(self, check):
                 outfile.write("\n")
         return analysis
 
+    def contains_summary_analysis(self):
+        """Return True if the summary file contains an analysis."""
+        with open(self.fullpath, "r") as file:
```
kastelecn: Is this not using FileLock by intense?

dhaumont: Nice catch ! indeed, I forgot to use the fileLock here. I'm fixing it right ahead

tactus/reference_checker.py
```diff
@@ -1171,8 +1205,19 @@ def execute(self, fmanager: FileManager):
...
-            self.create_summaries_with_header(fmanager.platform)
+        force_deletion = self.create_summary
+
+        for summary in self.summary_list:
+            if os.path.exists(summary.fullpath):
+                delete = force_deletion
+                if not delete:
+                    has_summary = summary.contains_summary_analysis()
+                    if has_summary:
+                        delete = not self.analyze_summary
```
kastelecn, line 1216: Is there right logic, don't we want to delete summary if we want to analyze again?

dhaumont: > Is there right logic, don't we want to delete summary if we want to analyze again?

The idea is that the summary file contains a list of results and a final analysis in the end.

The comparison against references is performed by the computation tasks themselves: they append the result of their comparisons to the summary file. The ReferenceCheck task, at the end of the suite, only reads the summary file - there is no comparison done at that point - and happens the final analysis section to it. It will also throw an exception if some tests are failing.

If you want to analyze again, you have to restart the suite (or the corresponding tasks), not only the ReferenceCheck task. Of course, you can then also relaunch the ReferenceCheck, to create the final analysis.

In the suggested implementation, we delete the summary under two conditions:
1. we are in the `Prep` task (self.create_summary == True) - this is the initial behaviour in develop
2. we are in a computation task but the summary file already contains the final 'analysis' section and we can´t happen anything to it anymore (`not self.analyze_summary and has_summary') - this is something new.

Note also that an alternative for case 2. would be to remove the analysis section from the summary file, happens the new comparions results to the summary file, and add back the analysis section. But the side effect is that the analysis section would contain results from previous run and from the new run, mixed togheter, so I think it's better to delete the summary file as implemented in this PR.

Does it answer your question ?

tactus/reference_checker.py
dhaumont: Solved in c471068b7c042979e0b86b88c1fe24917f897713

kastelecn (approved):
LGTM

PR #132, opened 2026-06-23
Add `tactus compile` subcommand for IAL compilation 

pardallio:
Add `tactus compile` subcommand for IAL compilation configurations, including the `cy50t2_compile` preset, `@IAL_TAG@` macro, and supporting documentation.

### Added
- New `tactus compile` subcommand for building IAL compilation configurations. Wraps `tactus case` with compilation-specific defaults (`@HOST@.toml` and `compile_suite.toml` modifications).
- `--ial-tag` / `-s` (start-suite) / `--case-name` / `--output` options on the `compile` subcommand.
- `create_compile_exp` function in `commands_functions.py` implementing the `compile` command logic.
- New configuration preset `tactus/data/config_files/configurations/cy50t2_compile` for CY50t2 IAL compilation.
- New `@IAL_TAG@` macro mapped to `compile.ial_git_branch` in `include/macros.toml`.
- Documentation: "Running a Compilation" section in `compiling.md` and "Compiling IAL via the `compile` command" section in `case_config.md`.

### Changed
- `compile_suite.toml` now sets a default case name (`IAL_@IAL_TAG@_compile`) under `[general]` and enables `suite_control.compile = true`.
## Type of change

### Testing

### Code quality

## Checklist for reviewers
- [ ] the code readable
- [ ] the code documented

  missing)

## Checklist for assignees
- [ ] the tests passing

## Tag possible reviewers


pardallio:
# Running a Compilation

Tactus exposes a dedicated `compile` subcommand that builds a configuration aimed at compiling IAL and starts (optionally) the compilation suite.

## Quick start

```
tactus compile --ial-tag develop -s
```

This will:

1. Build a config from `config.toml` +  host-specific overrides + `compile_suite.toml`
2. Set `compile.ial_git_branch` to `develop`
3. Generate a case named `IAL_develop_compile`
4. Start the compilation suite (`CompilationSuiteDefinition`) because of `-s` (equivalent to `--start-suite`)

## What the command does

The `compile` subcommand is a specialization of `tactus case`. It:

* Sets `compile.ial_git_branch` from `--ial-tag` (default: `develop`)
* Always merges the following modification files on top of the user-supplied config:
  * `tactus/data/config_files/modifications/@HOST@.toml`
  * `tactus/data/config_files/modifications/compile_suite.toml`
* Forwards everything else (output path, start-suite flag, keep-def-file, expand-config) to `tactus case`

dhaumont:
@pardallio Thank you for this proposal Miguel.
Could you add some unit testing, to check the syntax of your command (you can create a mockup for the compilation itself) ?

tactus/commands_functions.py
```diff
@@ -151,6 +151,24 @@ def create_exp(args, config):
         start_suite(args, config)
 
 
+def create_compile_exp(args, config):
+    """Implement the 'compile' command.
+
+    Args:
+        args (argparse.Namespace): Parsed command line arguments.
+        config (.config_parser.ParsedConfig): Parsed config file contents.
```
dhaumont, line 159: In the doc, rename .config_parser.ParsedConfig -> ParsedConfig

docs/markdown_docs/case_config.md
```diff
@@ -52,3 +52,30 @@ where `time.toml` contains
...
+* injects the requested IAL Git tag/branch into `compile.ial_git_branch`
+* uses the `CompilationSuiteDefinition` suite
+
+### Basic usage
+
+```
+tactus compile --ial-tag develop --start-suite
+```
+
+produces a nd launches case named `IAL_develop_compile`.
```
dhaumont: a nd -> and

docs/markdown_docs/compiling.md
```diff
@@ -448,3 +493,15 @@ exists, compilation is skipped. The cache symlink is still (re)created so the lo
...
 * The `ecbundle` binary is resolved as `<python-bin-dir>/ecbundle`, i.e. it must be installed in the same environment as Tactus
 * `IAL_DIR` is always exported from `compile.ial_dir`, regardless of whether `bundle_update` is enabled, so bundle YAMLs can rely on it being set
+
+
+
+```
+
+---
+
+A couple of things worth flagging honestly:
```
dhaumont: Is it AI generated comment ? Do we want it here ?

docs/markdown_docs/case_config.md
```diff
@@ -52,3 +52,30 @@ where `time.toml` contains
...
+
+* forces inclusion of the host-specific modifications (`@HOST@.toml`)
+* forces inclusion of the compilation suite modifications (`compile_suite.toml`)
+* injects the requested IAL Git tag/branch into `compile.ial_git_branch`
+* uses the `CompilationSuiteDefinition` suite
+
+### Basic usage
+
+```
+tactus compile --ial-tag develop --start-suite
```
dhaumont: 1. Why did you keep the  `--start-suite` argument ?

Can't we just do
```
tactus compile --ial-tag develop 
```

2. as `develop` is the default for  `-ial-tag`, I would show instead a command uising a real tag and/or a commit hash

tactus/argparse_wrapper.py
```diff
@@ -255,6 +256,51 @@ def get_args_parser(program_name=GeneralConstants.PACKAGE_NAME):
...
+    ###########################################
+    # Configure parser for the "compile" command #
+    ###########################################
+    parser_compile = subparsers.add_parser(
+        "compile",
+        help="Start a compilation suite",
+        parents=[common_parser],
+    )
+    parser_compile.add_argument(
+        "--output",
```
dhaumont, line 268: Why do we need `--output` and `--start-suite` ?
Wasn't the goal of the compile comnand to hide this to the user ?
Otherwise I don't really see any advantage in comparison to 'tactus case config.toml --start suite`

docs/markdown_docs/compiling.md
```diff
@@ -1,3 +1,48 @@
...
+* Always merges the following modification files on top of the user-supplied config:
+  * `tactus/data/config_files/modifications/@HOST@.toml`
+  * `tactus/data/config_files/modifications/compile_suite.toml`
+* Forwards everything else (output path, start-suite flag, keep-def-file, expand-config) to `tactus case`
+
+## Required configuration
+
+For the compilation suite to run end-to-end, the following keys are read by the task classes documented below:
+
+| Key | Used by | Notes |
```
dhaumont, line 32: Is this table something we provide for the other commands ? We should keep the doc as uniform  as possible, so I would not include it.

tests/unit/test_compile_command.py
```diff
@@ -0,0 +1,114 @@
...
+def test_create_compile_exp_uses_default_ial_tag_from_parser(default_config, monkeypatch):
+    """Check compile command uses the parser default IAL tag when none is provided."""
+    captured = {}
+
+    def fake_create_exp(args, config):
+        captured["args"] = args
+        captured["config"] = config
+
+    monkeypatch.setattr("tactus.commands_functions.create_exp", fake_create_exp)
+
```
dhaumont, line 99: There is a lot of empty lines in this file, is the linter happy with this ?

docs/markdown_docs/compiling.md
pardallio: I provide a similar table for the other sections in this document

docs/markdown_docs/case_config.md
pardallio: once we have a working tag we will update it

tactus/argparse_wrapper.py
pardallio: 1.  It seems to me that it is wise to not  have the suite launched automatically
2.  `output `is not required since it has a default value, but the user can change it if they so wish
3. `tactus compile --ial-tag asdfg --start-suite` is a lot less complicated that `tactus case ?/home/prt2342/Documents/Dev/tactus/tactus/data/config_files/configurations/cy50t2_compile --start-suite` and it does something different, since it updates the tag directly instead of asking the user to edit the config file themselves.

since this compile command is a wrapper arround the case command it will always need  `start-suite` to be defined, we can make it true by default but it does not seem the best option to me.

I am open to changing it of course, but maybe a more complete discussion is needed

tactus/commands_functions.py
pardallio: Here I copied it from the other commands functions (which are all like this), so I don't think this one should be different form the other

docs/markdown_docs/case_config.md
dhaumont: Can't you already put CY50_toT1.05 ? Otherwise I'm afraid we will forget later on

tactus/argparse_wrapper.py
dhaumont: Sorry, I don't understand what the workflow will look like.
If I undertand correctly:

1. this command won't do anything:

```
tactus compile --ial-tag asdfg
```

2. to start compilation, you have to do :
```
tactus compile --ial-tag asdfg --start-suite
```

3.  store a config file and start the compilation later on 

```
tactus compile --ial-tag asdfg -o compile_cy50t2_asdfg.toml
tactus compile -c compile_cy50t2_asdfg.toml --start-suite 
```

Instead, my proposal is to remove the `--start-suite` and have it implicit:

1. Run directly (start-suite is implicit)
```
tactus compile --ial-tag asdfg
```

2. Save and start later (start-suite deactivated when -o is provided, and implicit when -c is provided):

```
tactus compile --ial-tag asdfg -o compile_cy50t2_asdfg.toml
tactus compile -c compile_cy50t2_asdfg.toml 
```

tactus/commands_functions.py
dhaumont: No, it's the other way around.
If you grep "ParsedConfig", you will get > 100 hits, but grep ".config_parser.ParsedConfig" gives only 15 hits

tactus/argparse_wrapper.py
pardallio: it works exactly in the same way as `case`. 
you don't need the `-o` to output the config file, the `-o` is just for changing the output name of the config file from the default one.

In my opinion, it would be better to keep the behaviour consistent between  commands. 

`tactus compile --ial-tag asdfg --start-stuite`
or
`tactus compile --ial-tag asdfg -s`

OR

`tactus compile --ial-tag asdfg` (produces a config file named IAL_[git_tag]_compile.toml by default)
followed by
`tactus start suite -c IAL_asdfg_compile.toml`

tests/unit/test_compile_command.py
pardallio: apparently, yes

tactus/commands_functions.py
pardallio: This really is something that is not worth discussing, but all the doctoring in this file have `.config_parser.ParsedConfig`. It must have been a style decision made by the original author I don't think I will change all the others in this PR too.

pardallio: also there are much more instances of `tactus.ParsedConfig` in the other files (78 hits) than just `ParsedConfig` (5 hits). `.config_parser.ParsedConfig` gets 18 hits.

tactus/argparse_wrapper.py
dhaumont: @kastelecn @khintz @uandrae  @trygveasp 
What is your opinion regarding this discussion ? Any preference regarding the compile command we want to introduce ?

tactus/commands_functions.py
dhaumont: > This really is something that is not worth discussing, 

Quite the opposite, this is integral part of the review process to keep documentation and code consistent and this is the right place to have this kind of discussion.

I agree keep this consistent in the file in this PR, but we should take an action later on to have it unified accross tactus.

dhaumont: https://github.com/ACCORD-NWP/tactus/issues/133

tests/unit/test_compile_command.py
dhaumont: all the other tests have the same empty lines everywhere, so I will consider this is not an issue.

docs/markdown_docs/case_config.md
pardallio: Is that one working? https://github.com/ACCORD-NWP/IAL/releases/tag/CY50_toT1.05  is from more than 1 year ago

dhaumont: Oups, sorry. I was trying to refer to the latest tag of last week.

dhaumont:
@Miguel Did you already think about the usage of this feature in tactus test runner ?
Currently, ttr adds a compile suite in a test case for CY50t2, and the climate suites have a dependency on it

1. compile definition: https://github.com/pardallio/tactus/blob/030cc5c5449eabe3148d364eba73f758677ed715/tactus/data/tests/atos_bologna_CY50t2.toml#L4

2. dependency: https://github.com/pardallio/tactus/blob/030cc5c5449eabe3148d364eba73f758677ed715/tactus/data/tests/case_definitions_CY50t2.toml#L20 

I guess we can get rid of 1., but it's not clear how we handle the dependency with other suites. Any idea already ?

pardallio:
> @miguel Did you already think about the usage of this feature in tactus test runner ? Currently, ttr adds a compile suite in a test case for CY50t2, and the climate suites have a dependency on it
> 
> 1. compile definition: https://github.com/pardallio/tactus/blob/030cc5c5449eabe3148d364eba73f758677ed715/tactus/data/tests/atos_bologna_CY50t2.toml#L4
> 2. dependency: https://github.com/pardallio/tactus/blob/030cc5c5449eabe3148d364eba73f758677ed715/tactus/data/tests/case_definitions_CY50t2.toml#L20
> 
> I guess we can get rid of 1., but it's not clear how we handle the dependency with other suites. Any idea already ?

I don't think we can get rid of 1. , but the addition of  https://github.com/pardallio/tactus/blob/030cc5c5449eabe3148d364eba73f758677ed715/tactus/data/config_files/configurations/cy50t2_compile should instead allow for removing https://github.com/ACCORD-NWP/tactus/blob/5de8e7192e388ac92cdc8b0df4fdca7e84a6df16/tactus/data/tests/case_definitions_CY50t2.toml#L8-L16

dhaumont:
> I don't think we can get rid of 1.

This is very annoying, since if we keep 1, the compilation will still be started everytime the ttr tests are launched.
One of the main motivation to introduce the compilation command was to improve the ttr tests workflow , by using first the compilation command to prepare the executables, and then launch the tests, using the executables pre-builr.
Did I miss something ?

pardallio:
> > I don't think we can get rid of 1.
> 
> This is very annoying, since if we keep 1, the compilation will still be started everytime the ttr tests are launched. One of the main motivation to introduce the compilation command was to improve the ttr tests workflow , by using first the compilation command to prepare the executables, and then launch the tests, using the executables pre-builr. Did I miss something ?

I was thinking compilation could be a test itself, but maybe it doesn't make sense to enforce it every time

pardallio:
also, This discussion is maybe best kept for a different PR

dhaumont:
> also, This discussion is maybe best kept for a different PR

okay, I will investigate this further and create a PR

tactus/argparse_wrapper.py
```diff
@@ -255,6 +256,51 @@ def get_args_parser(program_name=GeneralConstants.PACKAGE_NAME):
...
+        "--case-name", dest="case", help="Case name", required=False, default=None
+    )
+    parser_compile.add_argument(
+        "--ial-tag",
+        dest="ial_tag",
+        help="IAL git tag/branch",
+        required=False,
+        default="develop",
+    )
+    add_keep_def_file(
```
dhaumont, line 298: Why do you need add_keep_def_file and add_expand_config ?

pardallio: because the `create_exp` routine needs it

tactus/argparse_wrapper.py
dhaumont: I tested and it works well, I will validate the PR when my last questions will be answered

dhaumont (approved):
LGTM

dhaumont:
> > also, This discussion is maybe best kept for a different PR
> 
> okay, I will investigate this further and create a PR

https://github.com/ACCORD-NWP/tactus/pull/134

pardallio:
> LGTM

I would merge this, but there is a change that I am not sure is 100% OK, https://github.com/ACCORD-NWP/tactus/pull/132/changes#diff-f9018647f65b3a0b787fd7e6b1f72b79b08cf88ef4f67f560c8e2cd7f77db03aR2

I change the default name of the compilation suite to` IAL_@IAL_TAG@_compile`, I don't know if this has implications on test runner

dhaumont:
> I change the default name of the compilation suite to` IAL_@IAL_TAG@_compile`, I don't know if this has implications on test runner

There is no impact on test runner, as `case` is redefined to `case = "@CASE_PREFIX@_@CYCLE@_compile"` in `tactus/data/tests/case_definitions_CY50t2.toml`:

In other words. by default,  `tactus compile`  creates a suite (based on the IAL tag name) which is not used by `tactus test` (looking for a suite based on tactus tag).

One way to connect both commands is to create the compilation suite like this, such that ttr finds it
```
  tactus compile -o <my_tactus_branch>_CY50t2_compile.toml
```

But as you suggested, we can adress this problem in a new PR:
https://github.com/ACCORD-NWP/tactus/pull/134/

tactus/argparse_wrapper.py
khintz: I would expect `tactus compile --ial-tag asdfg` to compile. It seems a bit redundant to me having to add `--start-suite`. So `tactus compile --ial-tag asdfg` doesn't do anything?

pardallio: It creates a config file that can be used to start a suite. It follows the same behaviour as `tactus case`.
 
In my opinion it would be better if the behaviour was consistent across the multiple tactus commands, but if multiple people agree that the default behaviour for the compile command should be to just start a suite then I can change it

kastelecn: I think that it is better to follows the same behavior as `tactus case`

dhaumont: Thanks for your feedback, @kastelecn and @khintz 

I suggest to keep it as it is, at least for a first version, we can always improve the behaviour later on.
LGTM !

uandrae:
In #145 I introduce the default to use precompiled binaries. Here I've introduced a structure like

```
acrd@ac6-102:/ec/project/accord/IAL$ tree -L 3
.
└── CY50t2
    ├── CY50T2_toT3.04
    │   └── R64
    ├── CY50T2_toT3.06
    │   ├── R32
    │   └── R64
    └── latest -> CY50T2_toT3.06
```
What's under R32/R64 is a direct copy of whats produced from this PR. Following this convention it would be natural to also add `@CYCLE@` in the `install/@CYCLE@/@PRECISION@` path in this PR. If we would like to future proof a bit we should perhaps also include `@COMPILER@`?

pardallio:
> In #145 I introduce the default to use precompiled binaries. Here I've introduced a structure like
> 
> ```
> acrd@ac6-102:/ec/project/accord/IAL$ tree -L 3
> .
> └── CY50t2
>     ├── CY50T2_toT3.04
>     │   └── R64
>     ├── CY50T2_toT3.06
>     │   ├── R32
>     │   └── R64
>     └── latest -> CY50T2_toT3.06
> ```
> 
> What's under R32/R64 is a direct copy of whats produced from this PR. Following this convention it would be natural to also add `@CYCLE@` in the `install/@CYCLE@/@PRECISION@` path in this PR. If we would like to future proof a bit we should perhaps also include `@COMPILER@`?

maybe a better option is the arch directory used in ecbundle like `ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0`. the `ecmwf/hpc2020` part may be redundant as its the host machine

uandrae:
> > In #145 I introduce the default to use precompiled binaries. Here I've introduced a structure like
> > ```
> > acrd@ac6-102:/ec/project/accord/IAL$ tree -L 3
> > .
> > └── CY50t2
> >     ├── CY50T2_toT3.04
> >     │   └── R64
> >     ├── CY50T2_toT3.06
> >     │   ├── R32
> >     │   └── R64
> >     └── latest -> CY50T2_toT3.06
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What's under R32/R64 is a direct copy of whats produced from this PR. Following this convention it would be natural to also add `@CYCLE@` in the `install/@CYCLE@/@PRECISION@` path in this PR. If we would like to future proof a bit we should perhaps also include `@COMPILER@`?
> 
> maybe a better option is the arch directory used in ecbundle like `ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0`. the `ecmwf/hpc2020` part may be redundant as its the host machine

So this would imply to include `@COMPILE.ARCH@` in the path, agree?

uandrae:
Following the discussion in our meeting I suggest to inlcude `--start-suite` as default.

pardallio:
> > > In #145 I introduce the default to use precompiled binaries. Here I've introduced a structure like
> > > ```
> > > acrd@ac6-102:/ec/project/accord/IAL$ tree -L 3
> > > .
> > > └── CY50t2
> > >     ├── CY50T2_toT3.04
> > >     │   └── R64
> > >     ├── CY50T2_toT3.06
> > >     │   ├── R32
> > >     │   └── R64
> > >     └── latest -> CY50T2_toT3.06
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > What's under R32/R64 is a direct copy of whats produced from this PR. Following this convention it would be natural to also add `@CYCLE@` in the `install/@CYCLE@/@PRECISION@` path in this PR. If we would like to future proof a bit we should perhaps also include `@COMPILER@`?
> > 
> > 
> > maybe a better option is the arch directory used in ecbundle like `ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0`. the `ecmwf/hpc2020` part may be redundant as its the host machine
> 
> So this would imply to include `@COMPILE.ARCH@` in the path, agree?


Yes, but that would be great, but it would be even better to have a 1 to 1 correspondence from the environment that is set on the submission section in the  tactus config file to the environment that is set on the arch folder.

pardallio:
> Following the discussion in our meeting I suggest to inlcude `--start-suite` as default.

so instead of `--start-suite`, the argument would be `--dry-run` to just produce the config file

uandrae:
> > > > In #145 I introduce the default to use precompiled binaries. Here I've introduced a structure like
> > > > ```
> > > > acrd@ac6-102:/ec/project/accord/IAL$ tree -L 3
> > > > .
> > > > └── CY50t2
> > > >     ├── CY50T2_toT3.04
> > > >     │   └── R64
> > > >     ├── CY50T2_toT3.06
> > > >     │   ├── R32
> > > >     │   └── R64
> > > >     └── latest -> CY50T2_toT3.06
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > What's under R32/R64 is a direct copy of whats produced from this PR. Following this convention it would be natural to also add `@CYCLE@` in the `install/@CYCLE@/@PRECISION@` path in this PR. If we would like to future proof a bit we should perhaps also include `@COMPILER@`?
> > > 
> > > 
> > > maybe a better option is the arch directory used in ecbundle like `ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0`. the `ecmwf/hpc2020` part may be redundant as its the host machine
> > 
> > 
> > So this would imply to include `@COMPILE.ARCH@` in the path, agree?
> 
> Yes, but that would be great, but it would be even better to have a 1 to 1 correspondence from the environment that is set on the submission section in the tactus config file to the environment that is set on the arch folder.

And how do we achieve this? I'm not familiar enough with all the build details to know what to look for.

uandrae (approved):
Happy to approve this and take the path-to-binaries discussion and changes in #145.

